PR #2, opened 2019-02-15
MarketAccess content changes / updating directory components

CWaszczuk-DIT:
Forgot to update the requirements to the correct version so that the header link will show on great. 
Also added some content changes that popped up today.


richtier:
@CJWaszczuk please edit the `.in` file then run `make compile_requirements`

https://github.com/uktrade/great-domestic-ui/blob/master/requirements.in#L21

that resolves versions and prevents version clashing

PR #5, opened 2019-02-19
TT-1076/TT-1082 updated slugs and added guidance for no verification code

depsiatwal:


depsiatwal:
screenshot 2019-02-19 at 18 38 22
screenshot 2019-02-19 at 19 16 07

contact/constants.py
```diff
@@ -23,6 +23,7 @@
 COMPANY_NOT_FOUND = 'company-not-found'
 VERIFICATION_CODE = 'verification-code'
 NO_VERIFICATION_LETTER = 'no-verification-letter'
+NO_VERIFICATION_MISSING = 'verification-missing'
```
richtier: oddly, all of these values seem to be present in https://github.com/uktrade/directory-constants/blob/master/directory_constants/constants/cms.py

could you use the values in https://github.com/uktrade/directory-constants/blob/master/directory_constants/constants/cms.py please

depsiatwal: These are used in RoutingFormView 
does that mean we want to change the pattern here ? to just refer to cms always e.g

 constants.NO_VERIFICATION_MISSING:
                build_great_account_guidance_url(
                cms.GREAT_HELP_VERIFICATION_CODE_MISSING_SLUG
            ),

richtier: just refer to cms.py sounds sensible :+1: 

Note sure why I didn't do that when I first added constants.py here

PR #11, opened 2019-02-27
Community join form page

erdem:


core/forms.py
```diff
@@ -1,5 +1,34 @@
...
+from django.core.validators import RegexValidator
+
+from core.constants import HEARD_ABOUT_CHOISES, COMPANY_SECTOR_CHOISES, EMPLOYEES_NUMBER_CHOISES
 
 
 class CompaniesHouseSearchForm(forms.Form):
     term = forms.CharField()
+
+
+class CommunityJoinForm(AbstractActionMixin, forms.Form):
```
richtier: if your goal is to send a gov notify email then you need:


```
from directory_forms_api_client.forms import GovNotifyActionMixin
from directory_components.forms import Form
from directory_components import fields

class CommunityJoinForm(GovNotifyActionMixin, forms.Form):
    name = fields.CharField(...)
    ...
```

note the user of directory_components's in order to solve the css style problems

see https://github.com/uktrade/great-domestic-ui/blob/master/contact/forms.py#L268 as an example

core/views.py
```diff
@@ -273,3 +274,13 @@ def get(self, request, *args, **kwargs):
         )
         api_response.raise_for_status()
         return JsonResponse(api_response.json()['items'], safe=False)
+
+
+class CommunityJoinFormPageCMS(FormView):
+    template_name = 'core/community-join-form.html'
+    form_class = CommunityJoinForm
+    success_url = reverse_lazy('community-form-success')
```
richtier: you will also need to trigger the sending of the confirmation email:
```
def form_valid(self, form):
        response = form.save(
            template_id=...,
            email_address=...,
            form_url=self.request.get_full_path(),
        )
        response.raise_for_status()
```

core/forms.py
erdem: Even if I used the same form class and view implementation, the form fields don't render correctly. The community form HTML elements didn't render with extra CSS classes. 

https://monosnap.com/file/HnVQ04e96Qfjjd4ZodtuqsCijh45Ij

I initialized the form view template the same as with the other views. Do you have any other idea why the form is not coming with CSS classes?

PS: The PR updated.

core/templates/core/includes/community_breadcrumbs.html
```diff
@@ -0,0 +1,15 @@
+<div class="container breadcrumbs">
+  <ol>
+    <li>
+      <a href="{% url 'landing-page' %}">great.gov.uk</a>
+    </li>
+    <li>
+      <a href="/community/">Community</a>
```
richtier: please use `{% url '...' %}`

erdem: The community page is not a permanent URL definition, created on cms. I can use `{% url wagtail_page hardcoded_community_page_slug %}` but it will be the same thing thought.

core/forms.py
richtier: replace `from django import forms`

with 
```
from directory_components.forms import Form
from directory_components import fields
```
and use directory_components Form and fields

core/forms.py
```diff
@@ -1,5 +1,100 @@
...
+                regex=r'^\+?1?\d{9,15}$',
+                message="Phone number must be entered in the format:"
+                        " '+999999999'. Up to 15 digits allowed."
+            )
+        ],
+        max_length=15,
+        min_length=8,
+        help_text=_('This can be a landline or mobile number'),
+        error_messages={
+            'max_length': _('Figures only, maximum 15 characters, minimum 8 characters excluding spaces'), # noqa
```
richtier: `noqa` is cheating :)

please make this multiline

core/templates/core/includes/community_breadcrumbs.html
richtier: as discussed - 

I sense a misunderstanding on what CMS is

CMS is just a data store of content

CMS users add content

other services retrieve the content (via an api endpoint call) and show it

great-domestic-ui just retrieves content from CMS and displays it in a page on great-domestic-ui (edited) 
I added the community url yesterday
https://github.com/uktrade/great-domestic-ui/blob/master/conf/urls.py#L514

core/forms.py
erdem: For me; it looks much readable without line breaks but I'll stay with the standard afterwards. Removed all of them. ;)

PR #14, opened 2019-03-01
XOT-663 SOO form mandatory fields

stevenburnell-dit:


contact/forms.py
```diff
@@ -372,28 +372,36 @@ class BusinessDetailsForm(forms.Form):
...
     )
     website_address = fields.CharField(
         label='Company website',
         help_text='Website address, where we can see your products online.',
         max_length=255,
-        required=False,
+        required=True,
     )
 
+    def clean(self):
```
sdonk: It'd be good to have a test for this piece of logic

stevenburnell-dit: Quite true. I believe @agiamas has it in mind as a nice introductory task for our new tester, @tapan-dit

sdonk: What I meant was a unittest in this code base, not an "external" test :)

PR #19, opened 2019-03-05
Added community section to domestic landing page

erdem:


article/templates/article/sections/community_section.html
```diff
@@ -0,0 +1,24 @@
...
+    <div class="grid-row">
+      <div class="column-half-l">
+        {% progressive_image webp='images/export-hp-community.webp' img_id='business-is-great-image' src='images/export-hp-communtiy.jpg' %}
+      </div>
+      <div class="column-full column-half-l">
+        <h2 id="business-is-great-title" class="heading-large">Join our export community</h2>
+        <p id="business-is-great-description">
+          Our export community is a network of businesses who advocate for exporting. They share success stories, offer practical advice and lead by example.
+        </p>
+		<br>
```
eadpearce: `br` tags are unnecessary here as `p`s force a line break by default and the extra spacing between paragraphs here is inconsistent with the rest of the page

erdem: I didn't want to put that `<br>` tags as well but it used like this in all other section templates. The `br` tags adding an extra margin between elements. That's the reason being people used it before. 

Besides, updating syles for `p` tags can be raised consistency problem on FE side. If we are cool with this, I am happy to clean `br` tags.

eadpearce: No what I'm saying is we don't want the extra spacing here on the homepage so remove `br` tags. No need to update `p` styling

erdem: Actually, the new design has extra spacing in that section. I spoke with the rest of the team, no needs extra spacing anymore. I removed the tags.

PR #24, opened 2019-03-06
Added nested fields to community join form

erdem:


community/tests/test_forms.py
```diff
@@ -20,7 +20,7 @@ def test_community_form_validations(valid_community_form_data):
 
 
 def test_community_form_api_serialization(
-        valid_community_form_data, serialized_community_form_api_data
+        valid_community_form_data
```
richtier: could be single line now the `serialized_community_form_api_data` is removed

PR #25, opened 2019-03-07
Tt 960 add nearby offices post code finder

depsiatwal:


contact/views.py
```diff
@@ -573,22 +577,26 @@ def flag(self):
...
+            address.append(office['address_city'])
+            address.append(office['address_postcode'])
+            office = {'address': '\n'.join(address), **office}
+            offices.append(office)
+        return offices if len(offices) > 0 else None
 
     def form_valid(self, form):
         office_details = self.format_office_details(form.office_details)
+        office_details = office_details[0] if office_details else None
+        other_offices = self.format_office_details(form.other_offices)
```
richtier: I suspect `format_office_details` exists to be DRY - but I think this is too DRY - the problem it's solving could simply be

```
other_offices = [office for office in office_list if not office['is_match']]
matching_offices = [office for office in office_list if office['is_match']]
office_details  = matching_offices[0] if matching_offices else None
````

and this function can be deleted

File: contact/forms.py [hunk 2]
```diff
@@ -493,16 +493,30 @@ class OfficeFinderForm(forms.Form):
     )
 
     @cached_property
-    def office_details(self):
+    def all_office_details(self):
```
richtier: this function was previously in the form because `helpers.retrieve_regional_offices` could return `400`, resulting in a validation error being raised by the form.

now there is no chance of 400 it would be better layering to move this logic from the form pleae

PR #27, opened 2019-03-08
Updated meta description logic in article pages

erdem:


article/templatetags/article_tags.py
```diff
@@ -10,3 +11,16 @@ def parse_date(date_string):
...
         return dateparser.parse(date_string).strftime('%d %B %Y')
     return None
+
+
+META_DESCRIPTION_TEXT_LENGTH = 150
+
+
+@register.simple_tag
+def get_meta_description(page, **kwargs):
+    description = page.get('article_teaser', '')
```
eadpearce: can you also add `seo_description` as a fallback here before we used the body text? it's another optional field in wagtail so it's often not filled out

PR #28, opened 2019-03-08
Feature/xot 655 in export readiness, build search results page back end

WillTaylor22:
Note:
The route /search here is not linked anywhere, but still would be accessible if someone randomly typed the url in. (It's not a broken path, so would not create an unduly poor experience - just a mostly blank page.)


core/templates/core/search.html
```diff
@@ -0,0 +1,37 @@
...
+{% block head_title %}Search - great.gov.uk{% endblock %}
+
+{% block content %}
+
+{% block breadcrumbs %}
+  <div class="container">
+    <div class="breadcrumbs body-text">
+      <ol>
+        <li>
+          <a href="/">great.gov.uk</a>
```
richtier: please use `{% url .. %}`

core/views.py
```diff
@@ -310,3 +315,100 @@ def form_valid(self, form):
...
+        request = requests.Request(
+            method=method, url=url, data=query).prepare()
+
+        auth = MohawkSender(
+            {
+                'id': settings.ACTIVITY_STREAM_API_ACCESS_KEY,
+                'key': settings.ACTIVITY_STREAM_API_SECRET_KEY,
+                'algorithm': 'sha256'
+            },
+            url,
```
richtier: why define the value outside of the function call? The following seems quicker for a dev to "parse" with their eyes :)

```
            url=settings.ACTIVITY_STREAM_API_URL,
            method="GET",
            content=query,
            content_type='application/json',
```

core/views.py
```diff
@@ -310,3 +315,100 @@ def form_valid(self, form):
...
+            )))
+
+        def sanitise_page(page):
+            return int(page) if int(page) > 0 else 0
+
+        def parse(results):
+            return json.loads(results.content)['orderedItems']
+
+        def flatten(results):
+            return list(map(lambda x: x['object'], results))
```
richtier: :heart: thin views

please move these elsewhere such as `helpers.py`. It looks like this is serializer-like functionality, so a class called `SearchSerializer` might even be nice :thinking: 

rule of thumb: if the code is not view-specific then chuck it out :)

makefile
```diff
@@ -119,10 +119,14 @@ TEST_SET_ENV_VARS := \
 	export CONTACT_EXPORTING_USER_NOTIFY_TEMPLATE_ID=debug; \
 	export CONTACT_EXPORTING_AGENT_SUBJECT=exporting-subject; \
 	export COMPANIES_HOUSE_API_KEY=debug; \
-	export CONTACT_ENQUIRIES_AGENT_EMAIL_ADDRESS=domesitc-enquiries@example.com
+	export CONTACT_ENQUIRIES_AGENT_EMAIL_ADDRESS=domesitc-enquiries@example.com; \
+	export ACTIVITY_STREAM_API_URL=http://localhost:8080/v1/; \
+  export ACTIVITY_STREAM_API_SECRET_KEY=incoming-some-secret-3; \
+  export ACTIVITY_STREAM_API_ACCESS_KEY=incoming-some-id-3; \
```
richtier: funy indent

richtier: funky indent

richtier: funy indent

makefile
```diff
@@ -119,10 +119,14 @@ TEST_SET_ENV_VARS := \
...
+  export ACTIVITY_STREAM_API_IP_WHITELIST=1.2.3.4,2.3.4.5
 
 debug_webserver:
-	$(DEBUG_SET_ENV_VARS) && $(DJANGO_WEBSERVER)
+	$(DEBUG_SET_ENV_VARS) && $(TEST_SET_ENV_VARS) && $(DJANGO_WEBSERVER)
```
richtier: `DEBUG_SET_ENV_VARS` and `TEST_SET_ENV_VARS` are purposefully seperated:

 - `DEBUG_SET_ENV_VARS` is devoid of secrets and environment-specific values, allowing `conf/.env` to work
- `TEST_SET_ENV_VARS` has placeholder values for secrets and environment-specific values, allowing tests to work seamlessly.

if we committed this change then devs would need to keep deleting values from `TEST_SET_ENV_VARS` in order to make their `conf/.env` file used.

core/views.py
WillTaylor22: url is used both in the request and in the header, however I do agree that it's better with the duplication so have changed it over.

core/templates/core/search.html
WillTaylor22: Changed!

PR #33, opened 2019-03-14
Feature/xot 655 search results page

WillTaylor22:


core/helpers.py
```diff
@@ -179,3 +183,110 @@ def search(cls, term):
...
+
+def format_query(query, page):
+    """ formats query for ElasticSearch
+    Note: ActivityStream not yet configured to recieve pagination,
+    will be corrected shortly. Hence commented-out lines.
+    """
+    RESULTS_PER_PAGE = 10
+    from_result = (page - 1) * RESULTS_PER_PAGE
+    return json.dumps({
+        "query": {
```
sdonk: Can you please not mix single and double quotes? 
I think the codebase mainly uses single quotes, it'd be better to have them consistent

core/templates/core/search.html
```diff
@@ -0,0 +1,114 @@
...
+    <h1 class="heading-xlarge">Search results for<span class=""> {{ query }}</span></h1>
+
+    <div class="actions">
+      {% if results %}
+        <p><a href="#search-results-list">{{ total_results }} results found<span class="verbose"> for food</span></a></p>
+      {% else %}
+        <p href="#search-results-list">No results found. Please try again</p>
+      {% endif %}
+
+      <form action="/search" class="search" method="get" aria-label="Search form">
```
sdonk: `{% url 'search' %}`

core/tests/test_helpers.py
```diff
@@ -230,3 +232,135 @@ def test_search_unauthorized():
...
+              ]
+          }
+        },
+        "from": 10,
+        "size": 10
+    })
+
+
+def test_search_with_activitystream():
+    ''' Simply check that it doesn't expload,
```
sdonk: Docstring triple quotes have to be double

core/tests/test_views.py
```diff
@@ -497,6 +499,118 @@ def test_about_view(client):
...
+                        'content':
+                        'Winter services for the properties1) Former...',
+                        'url': 'www.great.gov.uk/opportunities/2'
+                    }
+                }]
+            }
+        })
+        search.return_value = Mock(status_code=200, content=mock_results)
+
+        response = client.get("%s?q=services" % reverse('search'))
```
sdonk: `%` is the old style of string formatting, python now encourage the use of `format` of `f`, also `get` takes a `data` arg for the payload, please use that

core/helpers.py
```diff
@@ -179,3 +183,110 @@ def search(cls, term):
...
+
+
+def parse_results(response, query, page):
+    current_page = int(page)
+
+    content = json.loads(response.content)
+    results = list(map(lambda x: x['_source'], content['hits']['hits']))
+    total_results = content['hits']['total']
+
+    total_pages = int(ceil(total_results)/10)
```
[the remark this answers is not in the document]
WillTaylor22: It was returning a float! Probably the /10

core/helpers.py
WillTaylor22: Sure

core/helpers.py
```diff
@@ -179,3 +183,110 @@ def search(cls, term):
...
+def sanitise_query(query):
+    return " ".join(list(filter(
+        None, re.findall(r"[a-zA-Z0-9']*", query)
+    )))
+
+
+def sanitise_page(page):
+    if(page.isdigit() and int(page) > 0):
+        return int(page)
+    else:
```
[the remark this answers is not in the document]
WillTaylor22: Done

core/tests/test_helpers.py
WillTaylor22: Thx

core/templates/core/search.html
WillTaylor22: Done

core/tests/test_views.py
WillTaylor22: Done

core/helpers.py
```diff
@@ -179,3 +183,110 @@ def search(cls, term):
         else:
             url = cls.endpoints['search']
             return cls.get(url, params={'q': term})
+
+
+''' --- Search Helpers --- '''
+
+
+def sanitise_query(query):
+    return " ".join(list(filter(
```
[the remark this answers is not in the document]
WillTaylor22: Done

core/helpers.py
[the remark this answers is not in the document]
sdonk: Would `ceil(total_results/10)` work?

core/views.py
```diff
@@ -301,3 +303,34 @@ def form_valid(self, form):
 
 class BaseNotifyFormView(FormSessionMixin, SendNotifyMessagesMixin, FormView):
     pass
+
+
+class SearchView(TemplateView):
+    template_name = 'core/search.html'
+    ''' Search results page.
```
sdonk: Docstring takes double quotes and it goes under the class name

WillTaylor22: Done

core/templates/core/search.html
```diff
@@ -0,0 +1,114 @@
...
+<link href="{% static 'styles/pages/search-results.css' %}" rel="stylesheet" type="text/css" />
+{% endblock %}
+
+{% block head_title %}Search - great.gov.uk{% endblock %}
+{% block css_layout_class %}search-results-page{% endblock css_layout_class %}
+
+{% block content %}
+
+{% block breadcrumbs %}
+  <!--div class="container">
```
richtier: is this commented out on purpose? if so please delete it and re-add it once it's needed

core/templates/core/search.html
```diff
@@ -0,0 +1,114 @@
...
+        <p href="#search-results-list">No results found. Please try again</p>
+      {% endif %}
+
+      <form action="{% url 'search' %}" class="search" method="get" aria-label="Search form">
+        <label class="verbose" id="search-again-label" for="search-again-input">Search again</label>
+        <input autocomplete="off" class="text form-control" name="q" placeholder="" type="search" value="{{ query }}" id="search-again-input">
+        <input class="submit search-button" type="submit" value="Go">
+      </form>
+    </div>
+    <a class="back" href="#">Back</a>
```
richtier: where should this go? currently, it jumps to the top of the page

core/helpers.py
```diff
@@ -179,3 +183,110 @@ def search(cls, term):
         else:
             url = cls.endpoints['search']
             return cls.get(url, params={'q': term})
+
+
+''' --- Search Helpers --- '''
+
+
+def sanitise_query(query):
+    return " ".join(filter(
```
richtier: based on the tests, it looks like the purpose of this to avoid SQL injections?

It looks like the search provider is elasticsearch though, which does not use SQL.

if I'm correct - then this can be removed, no?

core/helpers.py
```diff
@@ -179,3 +183,110 @@ def search(cls, term):
...
+        return int(page) if int(page) > 0 else 1
+    except ValueError:
+        return 1
+
+
+def parse_results(response, query, page):
+    current_page = int(page)
+
+    content = json.loads(response.content)
+    results = list(map(lambda x: x['_source'], content['hits']['hits']))
```
richtier: this is fine, but it could be `results = [hit['_source'] for hit in content['hits']['hits'] ]`

core/helpers.py
```diff
@@ -179,3 +183,110 @@ def search(cls, term):
...
+    total_results = content['hits']['total']
+
+    total_pages = int(ceil(total_results/10.0))
+
+    prev_pages = list(range(1, current_page))[-3:]
+    if (len(prev_pages) > 0 and (prev_pages[0] > 2)):
+        show_first_page = True
+    else:
+        show_first_page = False
```
richtier: there's a built in django Paginator class that should solve this for you:

```
from django.core.paginator import EmptyPage, Paginator

        try:
            paginator = Paginator(range(total_results), 10)
            paginator_page = paginator.page(current_page)
        except EmptyPage:
            # do whatever you want when user request a page that cannot be served e.g,  page 2123213
```

then you can expose `'paginator_page': paginator_page` to the template and do something like
```
<span>Your search found {{ pagination.paginator.count }}
{% if pagination.has_previous %}
    <a href="?page={{ pagination.previous_page_number }}">Back</a>
{% endif %}

{% if pagination.has_next %}
    <a href="?page={{ pagination.next_page_number }}">Next</a>
{% endif %}
```

that would mean less code to write, test, read, and maintain :cool:

core/helpers.py
```diff
@@ -179,3 +183,110 @@ def search(cls, term):
...
+       'show_last_page': show_last_page
+    }
+
+
+def format_query(query, page):
+    """ formats query for ElasticSearch
+    Note: ActivityStream not yet configured to recieve pagination,
+    will be corrected shortly. Hence commented-out lines.
+    """
+    RESULTS_PER_PAGE = 10
```
richtier: `10` is also used to calculate how many pages there are in `parse_results`. Please move this outside of the function and reference `parse_results` in `parse_results` too

core/templates/core/search.html
```diff
@@ -0,0 +1,114 @@
...
+  </div>
+</section>
+
+{% if results %}
+  <section id="search-results-list">
+    <div class="container">
+      <h2 class="verbose">Results found</h2>
+      <p class="information verbose">Displaying items <b>1&nbsp;-&nbsp;25</b> of <b>{{ total_results }}</b> in total</p>
```
richtier: is `1 - 25` hard-coded on purpose?

core/views.py
```diff
@@ -301,3 +303,34 @@ def form_valid(self, form):
...
+            response = helpers.search_with_activitystream(elasticsearch_query)
+            if(response.status_code != 200):
+                return {
+                    'error_message': response.content,
+                    'error_status_code': response.status_code,
+                    'query': query
+                }
+            else:
+                return helpers.parse_results(response, query, page)
+        except ConnectionError:
```
richtier: please catch [`RequestExeptions`](http://docs.python-requests.org/en/master/_modules/requests/exceptions/) - which is the superclass of all exceptions that can occur when the request is made. this will make the error handling more robust.

please also "focus" the try..accept more:

```
try:
    response = helpers.search_with_activitystream(elasticsearch_query)
except RequestException:
    ...
else:
    if response.status_code != 200:
        ...
```

please also remove the overzelous brackets from simple if statements as that make it a bit harder to read

core/templates/core/search.html
WillTaylor22: Removed

core/templates/core/search.html
WillTaylor22: Asking designer - removed for now

core/helpers.py
WillTaylor22: Added

core/helpers.py
WillTaylor22: Replaced with new gist as you sent on slack

core/helpers.py
WillTaylor22: Thanks - Added to a list of tech debt, will be looking at later this week.

core/helpers.py
WillTaylor22: Done

core/templates/core/search.html
WillTaylor22: Solved

core/views.py
WillTaylor22: Added

PR #34, opened 2019-03-15
CMS-857 Accordions only display if it has minimum viable fields

Kenneth-Siu:


article/templates/article/components/accordions.html
```diff
@@ -1,7 +1,7 @@
 {% load add_export_elements_classes cta_link from directory_components_tags %}
 
 {% for accordion in accordions %}
-  {% if accordion.title %}
+  {% if accordion.title and accordion.teaser and accordion.num_of_subsections >= 2 and accordion.num_of_ctas == 3 %}
```
richtier: can you please move this logic to a helper function and add tests.

a simple `if else` is fine in a template, but 4 predicates is asking to be tested :)

richtier: note I'm not asking for a custom template tag - they suck :laughing:

Kenneth-Siu: Makes sense

PR #35, opened 2019-03-18
Template layout updates for community page

erdem:


core/sass/pages/_article-detail.scss
```diff
@@ -52,4 +52,7 @@
     }
   }
 
+  .community-article header{
+    padding-bottom:10px;
```
eadpearce: this is a very small change to apply to just one article. what's the reasoning behind this?

erdem: There is a space gap between two contents, the product people said needs to be lower space between them. 

check this SS:
https://monosnap.com/file/jvGkdBwfxlL4Av1fHDJpATPenUi3aK

article/templates/article/article_detail_content.html
```diff
@@ -40,6 +40,19 @@ <h2 class="heading-small">Tags</h2>
 
     <div class="grid-row">
       <div class="column-three-quarters">
+
+				<div class="grid-row">
+					<div class="column-seven-ninths article-content margin-top-0">
+						{% if page.article_teaser %}
+							<div class="text">
+								<p>
+									{{ page.article_teaser }}
```
eadpearce: this should go right after the article title and before the publish date here https://github.com/uktrade/great-domestic-ui/pull/35/files#diff-4ea4117b640929b4cec62dc6249c561fR17
like in the original version https://github.com/uktrade/great-domestic-ui/blob/b21a251f0cd08dd08184ec4bb899c8845be72b0f/prototype/templates/prototype/article_detail.html#L30

core/sass/pages/_article-detail.scss
eadpearce: ah i see now. no need for this. the teaser needs to be added under the article title. see my other comment

PR #43, opened 2019-03-20
Update to new header and footer + add services landing page

eadpearce:
Screen Shot 2019-03-26 at 12 46 23


article/tests/test_templates.py
```diff
@@ -112,7 +112,7 @@ def test_prototype_landing_page_news_section(mock_get_page, client, settings):
 def test_markets_link_in_header_when_feature_on(
     mock_get_page, client, settings
 ):
-    settings.FEATURE_FLAGS['MARKETS_PAGES_ON'] = True
+    settings.FEATURE_FLAGS['NEW_HEADER_FOOTER_ON'] = True
```
MichaelWalker: Do we also need to test the combination of these features?

ie, if the NEW_HEADER is off and the MARKETS_PAGES are on, I guess we would still want the market links to appear?
(perhaps its a little overkill)

conf/settings.py
```diff
@@ -279,6 +278,9 @@
 SECURE_HSTS_INCLUDE_SUBDOMAINS = True
 
 # HEADER/FOOTER URLS
+DIRECTORY_CONSTANTS_URL_GREAT_DOMESTIC = env.str(
+    'DIRECTORY_CONSTANTS_URL_GREAT_DOMESTIC', ''
+)
 DIRECTORY_CONSTANTS_URL_EXPORT_READINESS = env.str(
```
richtier: this can be removed now as `DIRECTORY_CONSTANTS_URL_GREAT_DOMESTIC` replaces it

makefile
```diff
@@ -84,14 +84,16 @@ DEBUG_SET_ENV_VARS := \
...
+	export FEATURE_MARKET_ACCESS_ENABLED=false; \
 	export FEATURE_MARKET_ACCESS_GOV_LINK_ENABLED=true; \
 	export FEATURE_LANDING_PAGE_EU_EXIT_BANNER_ENABLED=true; \
 	export COMMUNITY_ENQUIRIES_AGENT_EMAIL_ADDRESS=community@example.com \
 	export ACTIVITY_STREAM_API_URL=http://localhost:8080/v1/objects; \
 	export ACTIVITY_STREAM_API_SECRET_KEY=incoming-some-secret-3; \
 	export ACTIVITY_STREAM_API_ACCESS_KEY=incoming-some-id-3; \
-	export ACTIVITY_STREAM_API_IP_WHITELIST=1.2.3.4,2.3.4.5
+	export ACTIVITY_STREAM_API_IP_WHITELIST=1.2.3.4,2.3.4.5; \
+	export DIRECTORY_CONSTANTS_URL_GREAT_DOMESTIC=http://exred.trade.great:8007; \
```
richtier: please also remove DIRECTORY_CONSTANTS_URL_EXPORT_READINESS

article/tests/test_templates.py
richtier: the makefile has `FEATURE_MARKETS_PAGES_ENABLED=true` so the removed line was redundant really.

conf/settings.py
```diff
@@ -380,7 +382,7 @@
 
 # feature flags
 FEATURE_FLAGS = {
-    'EXPORT_JOURNEY_ON': env.bool('FEATURE_EXPORT_JOURNEY_ENABLED', True),
+    'NEW_HEADER_FOOTER_ON': env.bool('FEATURE_NEW_HEADER_FOOTER_ON', False),
```
richtier: for consistency with other env vars please use `env.bool('FEATURE_NEW_HEADER_FOOTER_ENABLED`, ...)

PR #45, opened 2019-03-21
Prefill using data saved in forms api

mkieblesz:


mkieblesz:
As discussed in https://github.com/uktrade/directory-forms-api/pull/40 and [XOT-632](https://uktrade.atlassian.net/browse/XOT-632), we are going to use formtools custom storage, instead of poll forms-api for submissions.

PR #47, opened 2019-03-21
Add TEST_SET_ENV_VARS to debug_webserver command

thibaudcolas:
> I'm not entirely sure whether that change is appropriate, or whether I just missed part of the setup.

Trying to `make debug_webserver` for the first time, I got the following error:

```
UserWarning: /Users/tcolas/Dev/dit/great-domestic-ui/conf/.env doesn't exist - if you're not configuring your environment separately, create one.
[...]
KeyError: 'COMPANIES_HOUSE_API_KEY'
```

There seems to be quite a few of those variables that are not defined in `DEBUG_SET_ENV_VARS`, but are in `TEST_SET_ENV_VARS`. The other `debug_*` commands include `DEBUG_SET_ENV_VARS`, so I thought it might be appropriate to include it here as well?

---

Alternatively, it seems like the server is trying to load configuration from a `.env` file – it's mentioned in the README in the Docker section, but not too clear whether that would be meant to contain the missing variables. If that's the way to go it would need some documentation.


makefile
```diff
@@ -130,7 +130,7 @@ TEST_SET_ENV_VARS := \
 	export ACTIVITY_STREAM_API_IP_WHITELIST=1.2.3.4,2.3.4.5
 
 debug_webserver:
-	$(DEBUG_SET_ENV_VARS) && $(DJANGO_WEBSERVER)
+	$(DEBUG_SET_ENV_VARS) && $(TEST_SET_ENV_VARS) && $(DJANGO_WEBSERVER)
```
richtier: this is an undesirable change - `TEST_SET_ENV_VARS` contains placeholder values that will prevent secrets stored in `conf/.env` from being used

richtier:
secrets such as API keys and environment specific configurations are placed in `conf/.env` - a file that is not added to version control.

there are no sane defaults to these values, so `TEST_SET_ENV_VARS` just uses placeholder values like `'debug'`. this is just so the tests can run, which do not need to talk to third party services anyway.

if we merged this, then devs would need to either update `TEST_SET_ENV_VARS` to include secrets and remember not to merge them. in the past this is more-or-less what we did, and resulted in every so often a dev accidentally committing secrets.

so please instead utilize `conf/.env` file, which is read in `conf/settings.py` 
```
import environ
...
env = environ.Env()
env.read_env()

```

thibaudcolas:
Thanks @richtier! Should I take a stab at writing some info about `conf/.env` in the README? I could see this behavior from `settings.py` but couldn't see it documented.

Additionally – should there be some kind of template for `.env` that people could copy/paste and start from when they get the project set up locally?

thibaudcolas:
Take two after feedback from Rich: #48.

PR #52, opened 2019-03-22
Update dependencies

richtier:
I removed the js tests because:
 - they were not run on ci
 - they were really just testing that objects existed on the page - not that they behave how we wanted

given the overhead of maintaining those dependencies, it's not worth keeping them.

fortunately, we have very little js to speak of.


package.json
```diff
@@ -10,15 +10,11 @@
...
-    "gulp": "^3.9.1",
-    "gulp-sass": "^3.1.0",
-    "gulp-sourcemaps": "^2.6.1",
-    "karma": "^2.0.0",
-    "karma-chai": "^0.1.0",
-    "karma-chrome-launcher": "^2.2.0",
-    "karma-mocha": "^1.3.0",
-    "mocha": "^4.1.0",
-    "del": "^3.0.0"
+    "chai": "^4.2.0",
```
eadpearce: still keeping chai?

eadpearce:
you can get rid of the test gulp task as well https://github.com/uktrade/great-domestic-ui/pull/52/files#diff-b9e12334e9eafd8341a6107dd98510c9R16

package.json
richtier: oopw

README.md
```diff
@@ -22,23 +21,23 @@ We aim to follow [GDS service standards](https://www.gov.uk/service-manual/servi
...
 
 ### SASS
 We use SASS CSS pre-compiler. If you're doing front-end work your local machine will also need the following dependencies:
 
 [node](https://nodejs.org/en/download/)
-
 [SASS](https://rubygems.org/gems/sass/versions/3.4.22)
 
+
 ### Run JS tests with gulp or npm
```
eadpearce: missed this before. you can also delete this bit

PR #57, opened 2019-03-25
[XOT-632] Django cache wizard form data storage

mkieblesz:
This implements:
* form data persists based on authenticated user, rather then on session cookie
* after successful submit form data is persisted so it can be reused in another market application


contact/views.py
```diff
@@ -465,8 +468,9 @@ def slug(self):
 
 
 class SellingOnlineOverseasFormView(
-    mixins.NotFoundOnDisabledFeature, mixins.PreventCaptchaRevalidationMixin,
-    FormSessionMixin, mixins.PrepopulateFormMixin, NamedUrlSessionWizardView
+    SSOLoginRequiredMixin, mixins.NotFoundOnDisabledFeature,
```
richtier: `SSOLoginRequiredMixin` prevents logged out users from accessing the page. is that a desired change?

contact/views.py
```diff
@@ -508,9 +512,28 @@ def get_form_kwargs(self, *args, **kwargs):
...
+    def get_cache_prefix(self):
+        return 'selling_online_overseas_form_view_{}'.format(
+            self.request.sso_user.id)
+
+    def get_form_data_cache(self):
+        return cache.get(self.get_cache_prefix(), None)
+
+    def set_form_data_cache(self, form_data):
+        cache.set(
+            self.get_cache_prefix(), form_data, SOO_SUBMISSION_CACHE_TIMEOUT
```
richtier: glad you remembered the timeout argument. last time I did `cache.set` I neglected that end the data was lost after like 5 mins :laughing:

contact/views.py
mkieblesz: Yes that's expected.

PR #58, opened 2019-03-25
Style office finder results

richtier:
image

[this task](https://uktrade.atlassian.net/browse/TT-1157)


contact/templates/contact/office-finder.html
```diff
@@ -36,32 +35,38 @@ <h3 class="heading-medium">{{ office_details.phone_other_comment }}</h3>
...
-					{% endif %}
-					<br>
-					<a href="{% url 'office-finder-contact' postcode=form.cleaned_data.postcode %}" class="button">Contact {{ office.name }} office</a>
-				</section>
+                    <div class="column-one-quarter">
+                        {{ office.address|linebreaks }}
+                    </div>
+                    <div class="column-one-third">
+                        <h3 class="heading-small">Telephone</h3>
+                        <p><a href="tel:{{ office.phone }}">{{ office.phone }}</a></p>
```
eadpearce: `link` class pls!

contact/templates/contact/office-finder.html
```diff
@@ -36,32 +35,38 @@ <h3 class="heading-medium">{{ office_details.phone_other_comment }}</h3>
...
                 <a href="{% url 'office-finder-contact' postcode=form.cleaned_data.postcode %}" class="button">Contact {{ office_details.name }} office</a>
             </section>
-            <section>
-                <a href="{% url 'office-finder' %}" class="previous-step">Back</a>
-            </section>
         {% endif %}
     </div>
 
-    <div class="container">
+    <section class="container" id="other-resullts">
```
Kenneth-Siu: Little typo?

contact/templates/contact/office-finder.html
```diff
@@ -36,32 +35,38 @@ <h3 class="heading-medium">{{ office_details.phone_other_comment }}</h3>
...
+                        {{ office.address|linebreaks }}
+                    </div>
+                    <div class="column-one-third">
+                        <h3 class="heading-small">Telephone</h3>
+                        <p><a href="tel:{{ office.phone }}">{{ office.phone }}</a></p>
+                        {% if office.phone_other %}
+                            <h3 class="heading-smal">{{ office.phone_other_comment }}</h3>
+                            <p><a href="tel:{{ office.phone_other }}">{{ office.phone_other }}</a></p>
+                        {% endif %}
+                        <a href="{% url 'office-finder-contact' postcode=form.cleaned_data.postcode %}" class="link">Contact {{ office.name }} office</a>
```
eadpearce: more of a content/accessibility thing: it's not clear from the text what contact method this link is going to use. since we have a phone link above it people may think this is just the same thing repeated. maybe the content should be something like "Contact {{ office.name }} office via an online form" ?

contact/templates/contact/office-finder.html
```diff
@@ -36,32 +35,38 @@ <h3 class="heading-medium">{{ office_details.phone_other_comment }}</h3>
...
 		{% if other_offices %}
-			<h2 class="heading-medium">List of other trade offices</h2>
-			{% for office in other_offices %}
-				<section class="lg-2" id="results">
-					<hr>
-					<h3 class="heading-medium" id="office-name">{{ office.name }}</h3>
-					{{ office.address|linebreaks }}
+			<div class="width-half">
+                <h2 class="heading-large">Department for International Trade regional offices</h2>
+            </div>
```
Kenneth-Siu: It looks like this file is a mix of tabs and spaces. (Leading to some really weird indentation depending on your tab settings...)

Does your IDE have a convenient way of enforcing one or the other?

core/sass/office-finder.scss
```diff
@@ -29,7 +43,26 @@ button {
...
   hr {
-    background-color: #969696;
+    background: $grey-2;
+  }
+}
+
+#other-resullts {
+  padding-bottom: 0;
+  h2 {
+    margin-bottom: 40px;
```
eadpearce: stick to the spacing scale pls. can this be 45px?

core/sass/office-finder.scss
```diff
@@ -29,7 +43,26 @@ button {
...
+    margin-top: 60px;
+  }
+  h3 {
+    margin-top: 0;
+    margin-bottom: 20px;
```
eadpearce: same here. can this be 15px or 30px?

Kenneth-Siu: No implication, when does it make sense to use the margin-top-xx CSS classes from directory-components over doing it manually in the SCSS file?

contact/templates/contact/office-finder.html
```diff
@@ -36,32 +35,38 @@ <h3 class="heading-medium">{{ office_details.phone_other_comment }}</h3>
...
+            {% for office in other_offices %}
+				<div class="grid-row">
+                    <h3 class="heading-medium" id="office-name">{{ office.name }}</h3>
```
eadpearce: we could put all these contact details inside an `address` element after the heading here rather than a `div` to add a bit more semantic-ness (headings are not allowed inside the `address` element) https://developer.mozilla.org/en-US/docs/Web/HTML/Element/address

core/sass/office-finder.scss
richtier: it's off plan according to the design but sure

core/sass/office-finder.scss
richtier: ah yes, I will use the utility classes :+1:

contact/templates/contact/office-finder.html
richtier: the first of many for me :laughing:

contact/templates/contact/office-finder.html
richtier: it tries to guess. I'll make all spaces

contact/templates/contact/office-finder.html
richtier: the use case for `<address>` seems ambiguous: be it for the contact details of our department - not arbitrary addresses, or be it for contact details in general?

the fact we cannot have headers in the address element would make it less semantic - we would need to change the h3s to e.g, spans which would overall be a regression in accessibility.

unless I'm missing something important, I think we can skip `<address>` here

eadpearce: yeah the documentation is not especially clear so i think it's fair to skip it

PR #62, opened 2019-03-28
Fix UI bug in office finder

richtier:
[this task](https://uktrade.atlassian.net/browse/TT-832)

image


richtier:
@Kenneth-Siu 
the reason the css was regenerated was because somehow I committed scss changes without committing the compiled css, causing the horizontal rule to have incorrect margin and colour.

compiling the css fixed it

PR #71, opened 2019-04-01
add key pages api

WillTaylor22:


conf/urls.py
```diff
@@ -212,12 +212,16 @@
         core.views.ServiceNoLongerAvailableView.as_view(),
         name='custom-page'
     ),
+    url(
+        r'^search/key-pages$',
```
sdonk: trailing slash missing

core/tests/test_views.py
```diff
@@ -610,6 +610,13 @@ def test_search_view(client):
         assert context['error_status_code'] == 500
 
 
+def test_search_key_pages_view(client):
+    response = client.get(reverse('search-key-pages'))
+    feed_parsed = json.loads(response.content)
+    assert feed_parsed["orderedItems"][0]["object"]["name"] == \
```
sdonk: please don't mix single and double quotes

core/views.py
```diff
@@ -328,5 +328,12 @@ def get_context_data(self, **kwargs):
                 return helpers.parse_results(response, query, page)
 
 
+class SearchKeyPagesView(TemplateView):
+    """ Returns data on key pages (such as the Get Finance homepage) to
+        include in search that are otherwise not provided via other APIs.
+    """
+    template_name = 'core/search-key-pages.txt'
```
richtier: `search-key-pages.txt` looks more like a json file. could you name it `search-key-pages.jsone, please

richtier: will services that talk to this endpoint care about the content-type header? if so the default content type will be `text/html` but for json responses it's normally `application/json`

WillTaylor22: It doesn't

PR #97, opened 2019-04-11
Importing to UK contact routing

richtier:
[this task](https://uktrade.atlassian.net/browse/TT-1324)


contact/views.py
```diff
@@ -467,35 +440,6 @@ class DefenceAndSecurityOrganisationFormView(
     )
 
 
-class InternationalSuccessView(CountryDisplayMixin, BaseSuccessView):
```
eadpearce: 👍🔥

PR #99, opened 2019-04-11
1264 services page template change

Kenneth-Siu:


conf/urls.py
```diff
@@ -219,7 +219,8 @@
         r'^search/$',
         activitystream.views.SearchView.as_view(),
         name='search'
-    )
+    ),
+    url(r'^i18n/', include('django.conf.urls.i18n')),
```
eadpearce: we don't use prefixed urls or the set_language url on here so you can remove this

PR #100, opened 2019-04-11
1270 add banner to markets page

Kenneth-Siu:


article/templates/article/markets_landing_page.html
```diff
@@ -1,18 +1,25 @@
 {% extends 'article/topic_list_base.html' %}
 
+{% load banner from directory_components_tags %}
+
 {% block child_pages %}
-  {% if page.teaser %}
+<section class="topic-list-section background-stone-30">
+  <div class="container padding-vertical-60">
+    {% banner banner_content='<p>There are more export market guides coming soon. In the meantime, if you have a question about an export market read the <a href="https://www.gov.uk/government/collections/exporting-country-guides">exporting guides</a> on GOV.UK, or <a href="https://www.great.gov.uk/contact/triage/location/">get in touch</a>.</p>' %}
```
eadpearce: might be easier to read if you just did this all in html rather than using the template tag? plus i think those govuk links might need to open in a new tab

PR #108, opened 2019-04-15
Cms 1246 accessibility bugs

Kenneth-Siu:


contact/templates/contact/routing/step-location.html
```diff
@@ -12,6 +12,6 @@ <h2 class="heading-large">Where are you based?</h2>
 
 {% block back_button %}
     {% if prev_url %}
-        <a href="{{ prev_url }}" class="previous-step">Back</button>
+        <a href="{{ prev_url }}" class="previous-step">Back</a>
```
kowalcj0: :+1:

article/templates/article/components/accordions.html
```diff
@@ -30,7 +30,7 @@
                 <div class="column-full">
                 {% endif %}
                   {% if subsection.icon %}
-                  <img class="accordion-icon" src="{{ subsection.icon.url }}"/>
+                  <img class="accordion-icon" src="{{ subsection.icon.url }}" alt="{{ subsection.heading }} icon"/>
```
kowalcj0: :+1:

core/static/js/dit.js
```diff
@@ -86,154 +88,8 @@ dit.utils = (new function () {
   }
 
 });
-// Responsive Functionality.
```
kowalcj0: @Kenneth-Siu 
A lot of (potentially useful) code was removed here. Will it affect responsiveness of our site in any way?

Kenneth-Siu: This code is a duplicate of ones that are

a) already in directory-components, and
b) only used for the headers (and therefore should only live in directory-components)

No responsiveness functionality has been removed. 🌞

PR #111, opened 2019-04-17
Removed urgency page and altered content

CWaszczuk-DIT:
User Research proved the urgency page caused more problems so it's been removed.
Content has also been updated to represent changes


conf/urls.py
```diff
@@ -598,11 +598,6 @@
         marketaccess.views.MarketAccessView.as_view(),
         name='market-access'
     ),
-    url(
-        r'^report-trade-barrier/report/emergency-details/$',
-        marketaccess.views.ReportBarrierEmergencyView.as_view(),
-        name='market-access-emergency'
-    ),
```
richtier: is there a sane place to redirect this to?

CWaszczuk-DIT: If this is a case of potential bookmarking, the service hasn't gone live yet so no-one would have used the pages outside of people testing.

PR #112, opened 2019-04-17
UKEF Pages

erdem:
The pages are implemented in this PR: 
- New top-level home page(`/get-finance/`)([XOT-811](https://uktrade.atlassian.net/browse/XOT-811))
- UKEF landing page(`/project-finance/`)([XOT-812](https://uktrade.atlassian.net/browse/XOT-812))
- UKEF Contact Form page(`/uk-export-contact-form/`) ([XOT-816](https://uktrade.atlassian.net/browse/XOT-816))
- UKEF Contact Form page success(`/uk-export-contact-form-success/`) ([XOT-818](https://uktrade.atlassian.net/browse/XOT-818))
- Country cover page(`/country-cover/`) ([XOT-817](https://uktrade.atlassian.net/browse/XOT-817))
- What we offer page [Article Detail] (`/what-we-offer-you/`) ([XOT-815](https://uktrade.atlassian.net/browse/XOT-815)) 
- How we assess your project [Article Detail] (`/how-we-assess-your-project/`) ([XOT-813](https://uktrade.atlassian.net/browse/XOT-813))


conf/urls.py
```diff
@@ -178,9 +179,9 @@
         name='casestudy-york-bag'
     ),
     url(
-        r"^get-finance/$",
+        r"^trade-finance/$",
```
erdem: The old `/get-finance/` page is the new trade finance page now.

conf/urls.py
```diff
@@ -637,10 +638,56 @@
     ),
 ]
 
+
+ukef_urls = [
+    url(
+        r"^get-finance/$",
+        ukef.views.HomeView.as_view(),
+        {'slug': 'get-finance'},
```
eadpearce: this is only used for CMS pages. if these don't use cms content there's no point setting the slug here

conf/urls.py
sdonk: Will this create 404s? If yes please add the relevant redirects

ukef/forms.py
```diff
@@ -0,0 +1,114 @@
...
+        }
+    )
+    business_name = fields.CharField(
+        label=_('Business name'),
+        max_length=50,
+        error_messages={
+            'required': _('Enter your business name'),
+        }
+    )
+    business_website = fields.CharField(
```
sdonk: URLField?

conf/urls.py
erdem: No, it doesn't necessary because we just swap the URL to two different pages.

erdem: https://github.com/uktrade/great-domestic-ui/pull/112/files#diff-43a35a60011b999b2fa2757114c1d24aR637

URL definition is here.

sdonk: I assume this link https://www.great.gov.uk/get-finance/ will break when the new url will be released, right? That needs a redirect

ukef/views.py
```diff
@@ -0,0 +1,58 @@
...
+from django.urls import reverse_lazy
+from django.views.generic import TemplateView
+
+from article.views import CMSPageView
+from core.helpers import NotifySettings
+from core.views import BaseNotifyFormView
+from ukef.forms import UKEFContactForm
+
+
+class HomeView(CMSPageView):
```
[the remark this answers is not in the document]
erdem: @eadpearce I got your point but I think, CMS pages necessary for the meta info for HTML pages. The editors could be used only meta tags and description updates to CMS dashboard. For the other parts, I agreed `CMSPageView` doesn't necessary for static content but for meta tags would be nice.

conf/urls.py
erdem: Same as I explained on the top. CMS base view implementation could be useful for meta info.

conf/urls.py
sdonk: I see, `trade-finance` doesn't exist yet. I was confused because the diff is not against `develop` so it was showing it as a change

ukef/forms.py
richtier: sometimes we want to use `CharField` for websites bacause URLField expect websites to start with http or https - which is a valid check if the URL is to be used on a web page - but if the url is just submitted to a service desk then it's not needed. @erdem to confirm

conf/urls.py
richtier: actually https://www.great.gov.uk/get-finance/ is live in prod, so we would want a redirect :)

image

ukef/forms.py
sdonk: I agree with that, I added the comment because the error_message includes http(s). Let's wait for @erdem to confirm

ukef/views.py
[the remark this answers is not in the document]
eadpearce: there's literally no point creating and requesting an entire CMS page just for two fields. either make the whole thing a CMS page or don't. just hardcode the meta like the rest of the content if it's static. having partial content from the CMS is incredibly confusing for devs and content editors

ukef/forms.py
erdem: @sdonk  The Django URLField is not a user-friendly solution, It always required to write `http` prefix otherwise it rasing validation error. it so annoying. Charfield could be a better solution but I'll look for a good regex validation for this.

conf/urls.py
eadpearce: @richtier @sdonk no this is correct. there would be no 404 since /get-finance url still exists. maybe the better question to ask is: are users who go to /get-finance (e.g. because they have it bookmarked) and then see different content going to be confused by this? if so, maybe these urls should be renamed. this is more of a service design/content question though

ukef/forms.py
sdonk: Charfield is fine, your error message is confusing though because it suggests you must use http

erdem: @sdonk didn't see you next messages, looks like @richtier beat my comment with 2 mins. :)

erdem:
@eadpearce @sdonk As we discussed with @richtier,  I'll create separated pull-requests for each ticket along with your feedback. (Feature flag also coming with the new pull-requests.)

PR #113, opened 2019-04-18
Feature/XOT-763 ga360 tagging

stevenburnell-dit:
~Two out of seven pages (Home and ServicesLanding)~ All pages done!


core/mixins.py
```diff
@@ -157,3 +151,16 @@ def guess_family_name(self):
...
+    ga360_page_type = None
+
+    def get_context_data(self, *args, **kwargs):
+        ga360 = None
+        if self.ga360_page_type:
+            ga360 = {'page_type': self.ga360_page_type}
+        return super().get_context_data(
+            ga360=ga360,
+            *args, **kwargs
+        )
```
richtier: looks great overall :+1: 

could we simplify this a bit to:
```
class GA360Mixin:
    ga360_payload = None

    def get_context_data(self, **kwargs):
        return super().get_context_data(ga360=self.ga360_payload, **kwargs)
```

and each view that uses this mixin will be slightly more verbose:

```
ServicesView(mixins.GA360Mixin, BreadcrumbsMixin, TemplateView):
    ga360_payload = {'page_type': 'ServicesLandingPage'}
```

fewer moving parts we need to keep in our heads and opens up the ability to pass other arbitrary data other than page type :)

eadpearce: i left like that thinking we may need to add more stuff other than page type in future but eh, this is simpler. we can always rewrite it later

richtier: ty

PR #117, opened 2019-04-24
UKEF Home page view with feature flag

erdem:
https://uktrade.atlassian.net/browse/XOT-811


core/mixins.py
```diff
@@ -49,6 +49,13 @@ def flag(self):
         return settings.FEATURE_FLAGS['MARKET_ACCESS_ON']
 
 
+class UKEFPagesFeatureFlagMixin(NotFoundOnDisabledFeature):
```
richtier: could you move this to `ukef/views.py` please - `core` is for common code (or small bits of code that do not yet have a definitive home)

PR #118, opened 2019-04-25
UKEF Home page

erdem:
https://uktrade.atlassian.net/browse/XOT-811


conf/urls.py
```diff
@@ -179,7 +180,7 @@
     ),
     url(
         r"^get-finance/$",
-        finance.views.GetFinanceView.as_view(),
+        finance.views.TradeFinanceView.as_view(),
```
erdem: @richtier  This URL will change to `trade-finance` after UKEF pages published. I changed view naming only for now.

ukef/templates/ukef/home_page.html
```diff
@@ -0,0 +1,138 @@
...
+								</ul>
+							</div>
+						</a>
+					</div>
+
+				</div>
+
+				<div class="column-half-l column-full">
+					<div class="card" id="large-corporations-card">
+						<a class="card-link" href="#" id="large-corporations-link">
```
richtier: is this href meant to be empty?

erdem: The empty links will populate next PR, the pages will be disabled on production.

PR #120, opened 2019-04-29
UKEF Contact Page

erdem:
https://uktrade.atlassian.net/browse/XOT-816

contact_form_ss


Kenneth-Siu:
Could you add `UKEF_CONTACT_AGENT_EMAIL_ADDRESS` to the makefile's `DEBUG_SET_ENV_VARS`?

core/sass/ukef-contact-page.scss
```diff
@@ -0,0 +1,25 @@
+.uk-export-contact-form{
+  #id_like_to_discuss_other-container{
+    position: relative;
+    margin-left: 10px;
+    top: -20px;
+    border-left: 4px solid #D7D7D7;
+    padding-left: 20px;
+    margin-bottom: 10px;
+
+    @media (max-width: 769px) {
```
Kenneth-Siu: Should be 768px. Our breakpoint is that 769px and higher is large devices, so max-width would have to be 768px. You can check out the results in the directory-components demo pages and inspecting the various elements. https://directory-components-dev.herokuapp.com/responsive/

Kenneth-Siu: Although speaking of devices, we should try and do our CSS mobile first: the base CSS is for mobile and we apply adjustments to larger devices. i.e.

```
#id_like_to_discuss_other-container {
    top: 0;

    @media (min-width: 769px) {
        top: -20px;
    }
}
```

core/sass/ukef-contact-page.scss
```diff
@@ -0,0 +1,25 @@
...
+      top: 0;
+    }
+  }
+  @media (min-width: 641px) {
+    .autocomplete__wrapper {
```
Kenneth-Siu: I'm not sure I know what this class does. I can't find it anywhere, but I wouldn't be surprised if it's added because of the forms thing or by some javascript somewhere?

core/sass/ukef-contact-page.scss
```diff
@@ -0,0 +1,25 @@
...
+      font-size: 19px;
+      line-height: 1.31579;
+      width: 50%;
+    }
+  }
+
+  .teaser-text{
+    font-size: 24px;
```
Kenneth-Siu: Use the directory-components font size class instead of setting it here manually.

core/sass/ukef-contact-page-success.scss
```diff
@@ -0,0 +1,28 @@
...
+      content: '✔';
+      color: white;
+      position: absolute;
+      left: -50px;
+      font-weight: bold;
+      font-size: 36px;
+      top: 0;
+      line-height: 2.5em;
+    }
+  }
```
Kenneth-Siu: Does this stuff look okay on mobile?

core/sass/ukef-contact-page-success.scss
```diff
@@ -0,0 +1,28 @@
...
+
+  .thank-you-section{
+    padding-bottom: 40px;
```
Kenneth-Siu: May as well just use padding-bottom-45.

I've been hearing Michal get annoyed at designs not using the pattern library, so this might be a good opportunity to do that.

core/sass/ukef-contact-page-success.scss
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  .list.list-bullet.list-bullet li:before{
+    background-color: #000000;
+    border-radius: 8px;
```
Kenneth-Siu: Should this not just use the default bullet points? Again, this sounds like somewhere Michal would want the designer to use the standard pattern library.

core/sass/ukef-contact-page.scss
```diff
@@ -0,0 +1,25 @@
+.uk-export-contact-form{
+  #id_like_to_discuss_other-container{
+    position: relative;
+    margin-left: 10px;
+    top: -20px;
+    border-left: 4px solid #D7D7D7;
+    padding-left: 20px;
+    margin-bottom: 10px;
```
Kenneth-Siu: I get the feeling that if we have to do so many exact overrides, something's gone wrong somewhere in the design process. How does it look when you use the default margins and paddings?

core/sass/ukef-contact-page.scss
erdem: I updated the media queries but I took the reference one of the directory-components CSS files for that `769px` breakpoint which can't remember.

core/sass/ukef-contact-page.scss
erdem: The component was an autocomplete, I missed this. Removed.

core/sass/ukef-contact-page.scss
erdem: removed

core/sass/ukef-contact-page-success.scss
erdem: yeah, it looks fine to me.

core/sass/ukef-contact-page-success.scss
erdem: changed 👍

core/sass/ukef-contact-page-success.scss
erdem: I am taking the design reference and that was the original styling for the bullets. Should I use the original styling here?

core/sass/ukef-contact-page.scss
erdem: I prefer to use default margins and paddings classes but this field template coming from Django forms. I need to override the form method to update the template. it is a generic field.

core/sass/ukef-contact-page.scss
```diff
@@ -2,24 +2,13 @@
...
     margin-left: 10px;
-    top: -20px;
+    top: 0;
     border-left: 4px solid #D7D7D7;
     padding-left: 20px;
     margin-bottom: 10px;
 
-    @media (max-width: 769px) {
-      top: 0;
+    @media (min-width: 768px) {
```
Kenneth-Siu: Just one final thing left: the min-width/max-width media queries are inclusive, so we'd have either one of:

- `max-width: 768px`
- `min-width: 769px`

So in this case we'd need `min-width: 769px`

PR #122, opened 2019-05-02
Changed content in market access flow

CWaszczuk-DIT:


marketaccess/tests/test_forms.py
```diff
@@ -5,11 +5,13 @@
 
 @pytest.fixture
 def about_form_data():
+    business_type = ("I’m an exporter or investor, or"
```
eadpearce: you need a space at the end of this line

marketaccess/tests/test_views.py
```diff
@@ -47,6 +47,8 @@ def test_form_submission(mock_zendesk_action, client):
     settings.FEATURE_FLAGS['MARKET_ACCESS_ON'] = True
     url_name = 'report-ma-barrier'
     view_name = 'report_market_access_barrier_form_view'
+    business_type = ("I’m an exporter or investor, or"
```
eadpearce: same again here

marketaccess/forms.py
```diff
@@ -96,9 +96,10 @@ class ProblemDetailsForm(forms.Form):
         }
     )
     location = fields.CharField(
-        label='Where are you trying to export to?',
+        label='Where are you trying to export to or invest in?',
         error_messages={
-            'required': 'Tell us where are you trying to export to'
+            'required': 'Tell us where are you trying to \
+            export to or invest in'
```
richtier: can you stick to a consistent multi-line handling please - below brackets were used, here slash was used.

brackets are more common across the codebase so I would stick to that.

more consistency = less we have to think about when reading later :)

```
'required': (
    'Tell us where are you trying to export to or invest in'
)

PR #127, opened 2019-05-07
[CMS-1410] Add redirect URLs for updated export advice articles

kaedroho:
Note: This is only required on staging for now.


ababic:
@sdonk Changelog now updated

PR #128, opened 2019-05-08
UKEF What we offer page

erdem:
https://uktrade.atlassian.net/browse/XOT-815


ukef/templates/ukef/what_we_offer.html
```diff
@@ -0,0 +1,128 @@
+{% extends 'article/base.html' %}
+
+{% load static %}
+{% load article_tags %}
+{% load add_export_elements_classes from directory_components_tags %}
+
+{% block head_title %}What we offer you - great.gov.uk{% endblock %}
+
+{% block og_image %}{% if page.article_image %}{{ page.article_image.url }}{% else %}{{ block.super }}{% endif %}{% endblock %}
```
eadpearce: no point using `page.whatever` here as there's no cms page. just use the hero image url  from further down the template

ukef/templates/ukef/what_we_offer.html
```diff
@@ -0,0 +1,128 @@
...
+
+				<div class="grid-row">
+					<div class="column-three-quarters">
+						<img src="{% static 'images/what-we-offer-you-hero.png' %}" alt="" class="width-100">
+
+						<section class="grid-row">
+							<div class="column-seven-ninths article-content padding-bottom-0">
+								<h3 class="heading-medium">Project financing</h3>
+								<p class="text">
+									Through our <a href="https://www.gov.uk/guidance/buyer-credit-facility">Buyer Credit</a> and <a
```
eadpearce: please add `link` class to links

ukef/templates/ukef/what_we_offer.html
```diff
@@ -0,0 +1,128 @@
...
+					</div>
+				</header>
+
+				<div class="grid-row">
+					<div class="column-three-quarters">
+						<img src="{% static 'images/what-we-offer-you-hero.png' %}" alt="" class="width-100">
+
+						<section class="grid-row">
+							<div class="column-seven-ninths article-content padding-bottom-0">
+								<h3 class="heading-medium">Project financing</h3>
```
eadpearce: these follow an `h1` so they should be `h2`s rather than `h3`s

ukef/templates/ukef/what_we_offer.html
```diff
@@ -0,0 +1,128 @@
...
+								<h3 class="heading-medium">UKEF premiums</h3>
+								<p class="text">We can give you an <a href="https://www.gov.uk/government/publications/sovereign-premium-indicator-tool">indicative premium rate</a> for medium or long-term transactions, where we will be supporting an overseas sovereign buyer as the risk entity.</p>
+							</div>
+						</section>
+					</div>
+					<div class="column-quarter">
+						<div class="panel-vertical-narrow">
+							<ul>
+								<li>
+									<h3 class="heading-small margin-top-0"><a href="{% url 'how-we-assess-your-project' %}" class="link">How we assess your project</a></h3>
```
eadpearce: no point in these being headings. just apply the same classes to the `li` or the `a` instead

ukef/templates/ukef/what_we_offer.html
```diff
@@ -0,0 +1,128 @@
...
+	</section>
+
+	<section id="contact-us-section" class="contact-us-section padding-vertical-60 background-stone-30">
+		<div class="container">
+			<div class="grid-row">
+				<div class="column-three-quarters">
+					<div class="grid-row">
+						<div class="column-seven-ninths padding-0">
+							<div class="column-full padding-bottom-30">
+								<p class="text small">Contact us about your project.</p>
```
eadpearce: feels like this text should be a heading or at least styled to be bigger

eadpearce:
this text overlaps the image on tablet. pls fix
Screenshot 2019-05-08 at 10 14 16

ukef/templates/ukef/what_we_offer.html
```diff
@@ -0,0 +1,128 @@
...
+									</div>
+								</div>
+							</div>
+						</div>
+					</div>
+				</header>
+
+				<div class="grid-row">
+					<div class="column-three-quarters">
+						<img src="{% static 'images/what-we-offer-you-hero.png' %}" alt="" class="width-100">
```
eadpearce: ah i see what the problem is with the image. the class to use is `width-full` not `width-100`

ukef/templates/ukef/what_we_offer.html
erdem: nope, it is the original design font size.

ukef/templates/ukef/what_we_offer.html
erdem: I removed this block for now. The final content doesn't ready yet.

ukef/templates/ukef/what_we_offer.html
eadpearce: ah I just checked and we do use the same thing on the current get finance page. I think maybe it just looks odd because the content in this one is shorter

ukef/templates/ukef/what_we_offer.html
```diff
@@ -6,123 +6,121 @@
...
-					<li aria-current="page">Get in touch</li>
-				</ol>
-			</nav>
-		</div>
-	</section>
+  <section id="breadcrumbs-section" class="breadcrumbs-section">
+    <div class="container">
+      <nav class="breadcrumbs" aria-label="breadcrumb" role="navigation">
+        <ol>
+          <li><a href="/" class="link">Great.gov.uk </a></li>
```
eadpearce: link class isn't needed on breadcrumbs as the link styling comes from the `breadcrumb` class (this is pretty much the only place it isn't needed though)

ukef/templates/ukef/what_we_offer.html
```diff
@@ -6,123 +6,121 @@
...
-								</li>
-							</ul>
-						</div>
-					</div>
-				</div>
-			</div>
-		</article>
-	</section>
+                <h2 class="heading-medium">Insurance for UK exporters</h2>
+                <p class="text">
```
eadpearce: don't think the `text` class is needed on these `p` tags but it's not a big deal

ukef/templates/ukef/what_we_offer.html
```diff
@@ -6,123 +6,121 @@
...
+                  If you’re working in challenging markets, our insurance products let you manage risk, where the private market is unable to step in.
+                  <a href="https://www.gov.uk/guidance/export-insurance-policy" class="link">UKEF’s Export Insurance Policy</a> and
+                  <a href="https://www.gov.uk/guidance/bond-insurance-policy" class="link">Bond Insurance Policy</a> cover you in the event of
+                  non-payment or demand for payment under a bond, which is either unfair or caused by political events.
+                </p>
+                <h2 class="heading-medium">UKEF premiums</h2>
+                <p class="text">We can give you an <a href="https://www.gov.uk/government/publications/sovereign-premium-indicator-tool" class="link">indicative premium rate</a> for medium or long-term transactions, where we will be supporting an overseas sovereign buyer as the risk entity.</p>
```
eadpearce: sorry i forgot - for external links we usually open in a new tab so you'll need to add `rel="noopener noreferrer" target="_blank"`

PR #135, opened 2019-05-16
Bug/xot 830 events search results have incorrect domain

WillTaylor22:


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 - [[CMS-1271]](https://uktrade.atlassian.net/browse/CMS-1271) Update teaser section on country guide pages
 - [[MA-745]](https://uktrade.atlassian.net/browse/MA-745)Added Market Access tile to services pages
 - [[CMS-1410]](https://uktrade.atlassian.net/browse/CMS-1410) Added redirect URLs for updated export advice articles
+- [[XOT-830]](https://uktrade.atlassian.net/secure/RapidBoard.jspa?rapidView=73&modal=detail&selectedIssue=XOT-830) Parse Event URLS to user facing urls
```
richtier: security concerns were raised about linking to jira (many bits of small information can help hackers...erm...hack)

so please just include the ticket number see https://uktrade.atlassian.net/wiki/spaces/ED/pages/825786369/Changelog for more details

richtier: if you have the time could you also remove the existing links to jira from this changelog, please

activitystream/serializers.py
```diff
@@ -1,8 +1,25 @@
...
 from bs4 import BeautifulSoup
+from urllib3.util import parse_url
+
+from django.conf import settings
 
 
 def parse_search_results(content):
+
+    def strip_html(result):
+        content = result['content'] if 'content' in result else ''
```
richtier: `content = result.get('content', '')`

CHANGELOG.md
WillTaylor22: Will force push this away

WillTaylor22: `if you have the time could you also remove the existing links to jira from this changelog, please`
Sure; In this case there's no point in force pushing as it will be in the history anyway, so I shan't.

activitystream/serializers.py
WillTaylor22: tx

PR #138, opened 2019-05-17
Updated report a trade barrier form structure and content

CWaszczuk-DIT:


marketaccess/forms.py
```diff
@@ -87,20 +87,20 @@ class ProblemDetailsForm(forms.Form):
...
 
-    product_service = fields.CharField(
-        label='What goods or services do you want to export?',
-        help_text='Or tell us about an investment you want to make',
+    location = fields.CharField(
+        label='Where are you trying to export to or invest in?',
         error_messages={
             'required': (
-                'Tell us what you’re trying to export or invest in'
+                'Tell us where are you trying to export to or invest in'
```
eadpearce: instead of `where are you` shouldn't this be `where you are`?

marketaccess/forms.py
```diff
@@ -87,20 +87,20 @@ class ProblemDetailsForm(forms.Form):
...
     )
-    location = fields.CharField(
-        label='Where are you trying to export to or invest in?',
+    product_service = fields.CharField(
+        label='What goods or services do you want to export?',
+        help_text='Or tell us about an investment you want to make',
         error_messages={
             'required': (
-                'Tell us where are you trying to export to or invest in'
+                'Tell us what you’re trying to export or invest in'
```
erdem: it could be useful in the future if you are used Django translation utility function here. (`ugettext_lazy`)

PR #170, opened 2019-06-06
TT-1450 - change h2 to h1

depsiatwal:


depsiatwal:
No noticeable difference by change tags from H2 to H1 see attached..

image

PR #180, opened 2019-06-17
Update GA360 tagging

eadpearce:


core/mixins.py
```diff
@@ -153,11 +148,51 @@ def guess_family_name(self):
...
+
+        page_id = self.page['page_type']
+
+        breadcrumbs = self.page['tree_based_breadcrumbs']
+        site_subsection = self.page['title']
+        site_section = ''
+
+        # get the page one level up from the current page
+        if len(breadcrumbs) > 1:
+            site_section = breadcrumbs[-2]['title']
```
MichaelWalker: This doesn't quite match my understanding of site_section and site_subsection. (although there is a good change that I'm the one that's wrong! :) )

If we've got a structure like this

1) home_page/
2) home_page/industries/
3) home_page/industries/energy/
4) home_page/industries/energy/contact
then I'd expect the following
- they should all have the same 'business_unit' defined by the thing that the home_page is a home page off.
- the final 3 have the same 'site_section' of 'industries'.
- the final 2 should share a 'site_subsection' of 'energy'.
- 2) could have either a blank site_subsection or a 'listing' site_subsection or something.

This way our site_section / site_subsection is defined 'top down', and reflects the hierarchy of our site quite nicely, and so it makes it easy for us to see all 'industries' pages for example.

Using your system above I think we'd get a different approach of defining it 'bottom_up'. It's a little easier to code, but I don't think it gives us the same ability to categorise sections of our site.

(Or perhaps I'm missing something!)

core/mixins.py
```diff
@@ -153,11 +148,51 @@ def guess_family_name(self):
...
+
+        self.set_ga360_payload(
```
richtier: no need to unpack the dict into variables - could be something like
```
        self.set_ga360_payload(
            page_id=self.page['page_type'],
            business_unit=settings.GA360_BUSINESS_UNIT,
            site_section=breadcrumbs[-2]['title'] if len(breadcrumbs) > 1 else '',
            site_subsection=self.page['title']
        )
```
slightly easier to read

article/templates/article/components/cms_breadcrumbs.html
```diff
@@ -0,0 +1,21 @@
+
+{# For use with CMS pages whose urls reflect the tree structure in the CMS i.e. advice and markets pages #}
+
+<div class="container">
```
MichaelWalker: Nice! Could we have this in `directory_components` please? I'd like to start using it on international!

(I guess we'd have to specify the landing page, but everything else should just work?)

core/mixins.py
eadpearce: Yeah I tried to get `site_section` and `site_subsection` to fit `pageCategory` and `articleList` that were given in the spreadsheet so I didn't have to rewrite the middleware. It doesn't really work in domestic because the structure is different so I made it fit best I could. Unforch it ended up confusing.

eadpearce: As discussed I'll update this

core/mixins.py
eadpearce: haha i actually find that harder to read than what i did :'D

article/templates/article/components/cms_breadcrumbs.html
eadpearce: oh absolutely! yeah i've been thinking of adding homepage breadcrumbs labels/urls for each service to constants so we could do this

PR #193, opened 2019-06-21
[CMS-1666] Update path used to identify homepage

ababic:
What was previously the `ExportReadinessApp` page has been replaced with a `HomePage` (now applied to all environments), but keeping the original page slug. The old home page will be deleted from CMS soon, so this update is to make the front-end request the new version instead.

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -38,7 +38,7 @@ class LandingPageView(mixins.SetGA360ValuesForCMSPageMixin, TemplateView):
     @cached_property
     def page(self):
         response = cms_api_client.lookup_by_slug(
-            slug=slugs.GREAT_HOME,
+            slug='export-readiness-app',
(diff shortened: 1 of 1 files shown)


core/views.py
```diff
@@ -38,7 +38,7 @@ class LandingPageView(mixins.SetGA360ValuesForCMSPageMixin, TemplateView):
     @cached_property
     def page(self):
         response = cms_api_client.lookup_by_slug(
-            slug=slugs.GREAT_HOME,
```
MichaelWalker: Is it better to do this by updating the value of `slugs.GREAT_HOME`, or is that still a relevant value for something else?

ababic: I started doing that first, but then I wondered about the impact, as I don't know all of the places where that might be in use

PR #195, opened 2019-06-24
TT-1579 - handle invalid postcodes

depsiatwal:


contact/forms.py
```diff
@@ -299,8 +299,12 @@ def serialized_data(self):
...
             data['dit_regional_office_email'] = ''
         else:
-            data['dit_regional_office_name'] = details['name']
-            data['dit_regional_office_email'] = details['email']
+            if details:
+                data['dit_regional_office_name'] = details['name']
+                data['dit_regional_office_email'] = details['email']
+            else:
+                data['dit_regional_office_name'] = ''
+                data['dit_regional_office_email'] = ''
```
richtier: as this is done in two places could you instead do:

```
try:
    details = helpers.retrieve_regional_office(data['postcode'])
except requests.exceptions.RequestException:
    pass
else:
    data['dit_regional_office_name'] = details['name']
    data['dit_regional_office_email'] = details['email']
data.setdefault('dit_regional_office_name', '')
data.setdefault('dit_regional_office_email', '')
return data

```

depsiatwal: I have to check if details since API can now return empty rather then 404 when postcode not found this is causing contact us to crash

PR #204, opened 2019-06-28
Marketing article page template

erdem:
https://uktrade.atlassian.net/browse/XOT-922

screencapture-localhost-8007-campaigns-test-campaign-2019-06-28-18_04_56

### DIFF
--- a/article/templates/article/marketing_article_detail_content.html
+++ b/article/templates/article/marketing_article_detail_content.html
@@ -1,0 +1,93 @@
+{% load parse_date from article_tags %}
+{% load add_export_elements_classes from directory_components %}
+
+{% block breadcrumbs %}
+  {% include 'core/includes/breadcrumb.html' with breadcrumbs=breadcrumbs %}
+{% endblock %}
+
+<article class="article-main {{ view.slug }}-article">
+  <div class="container">
+
+    <header>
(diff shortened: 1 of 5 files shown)


article/templates/article/marketing_article_detail_content.html
```diff
@@ -0,0 +1,93 @@
+{% load parse_date from article_tags %}
+{% load add_export_elements_classes from directory_components %}
+
+{% block breadcrumbs %}
+  {% include 'core/includes/breadcrumb.html' with breadcrumbs=breadcrumbs %}
+{% endblock %}
+
+<article class="article-main {{ view.slug }}-article">
```
richtier: please avoid using view properties in the template - that relies on the template "knowing" about the view.

instead `slug` can be passed into the template via `get_context_data` on the view.

erdem: I updated but there was the same usage in default article template. I copied from that template.

erdem:
@richtier The PR description updated.

eadpearce:
From the screenshot this looks almost exactly like the article page we already have. Instead of adding a whole new template I think the current article should be rewritten in order to add overrides only for the parts that need modifying. This would save a lot of duplication of code

PR #237, opened 2019-07-30
XOT-968 - Add security fixes

WillTaylor22:


conf/settings.py
```diff
@@ -143,6 +143,8 @@
 # Django's default value for LANGUAGE_COOKIE_DOMAIN is None
 LANGUAGE_COOKIE_DOMAIN = env.str('LANGUAGE_COOKIE_DOMAIN', None)
 
+CSRF_COOKIE_HTTPONLY = True
```
sdonk: This is already enabled https://github.com/uktrade/great-domestic-ui/blob/develop/conf/settings.py#L325

finance/views.py
```diff
@@ -22,6 +23,7 @@ class TradeFinanceView(
     slug = slugs.GREAT_GET_FINANCE
 
 
+@method_decorator(never_cache, name='dispatch')
```
sdonk: What's the reason not to cache this view?

eadpearce: why is this needed?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - XOT-977 - Add marketing campaigns form, original clone of community join-form
 - No ticket - Consistent use of breadcrumbs component
 - TT-1678 - Rename EU exit to Brexit
+- XOT-968 - Fix security issues raised by latest pen test
```
sdonk: Because of the recent changelog update this is not in the incorrect place

conf/settings.py
WillTaylor22: HTTPONLY and SECURE are different... are you sure this is redundant?

richtier: confirmed not redundant :) https://docs.djangoproject.com/en/2.2/ref/settings/#csrf-cookie-httponly

WillTaylor22: Done thanks

PR #246, opened 2019-08-01
(add) in memory redirect URLs

agiamas:

 - [X] Changelog entry added.

### DIFF
--- a/core/middleware.py
+++ b/core/middleware.py
@@ -14,13 +14,8 @@ class CoreMiddleware(object):
 
     def process_request(self, request):
         url = request.get_full_path().split('/')[1]
-        print(url)
         if url and url != '':
             response = api_client.redirects.lookup_redirect_by_url(source_url=url)
-            response_id = response['id']
-            if response_id != None and response_id != '':
+            if response != None and 'id' in response.keys():
                 return redirect(response['target_url'], permanent=False)
         pass
-
-
-
--- a/core/middleware.py
+++ b/core/middleware.py
@@ -10,12 +10,12 @@ from django.conf import settings
 logger = logging.getLogger(__name__)
 
 
-class CoreMiddleware(object):
+class RedirectMiddleware(object):
 
     def process_request(self, request):
-        url = request.get_full_path().split('/')[1]
+        url = request.path
         if url and url != '':
-            response = api_client.redirects.lookup_redirect_by_url(source_url=url)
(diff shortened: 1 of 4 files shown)


core/middleware.py
```diff
@@ -0,0 +1,21 @@
...
+from django.shortcuts import redirect
+from django.http import HttpResponseNotFound
+from django.conf import settings
+
+
+
+logger = logging.getLogger(__name__)
+
+
+class CoreMiddleware(object):
```
richtier: `RedirectMiddleware` would be more apt

core/middleware.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def process_request(self, request):
+        url = request.get_full_path().split('/')[1]
```
richtier: [`request.path`](https://docs.djangoproject.com/en/2.2/ref/request-response/#django.http.HttpRequest.path)

core/middleware.py
```diff
@@ -0,0 +1,21 @@
...
+        if url and url != '':
+            response = api_client.redirects.lookup_redirect_by_url(source_url=url)
```
richtier: we could simplify this to `api_client.redirects.lookup_by_url(url)`

- only one argument, so we don't lose any readability by using arguments instead of keyword arguments
- it's already in the `redirects` namespace, so don't need to mention `redirects` in the method.

core/middleware.py
```diff
@@ -0,0 +1,21 @@
...
+            if response != None and 'id' in response.keys():
```
richtier: I think this should be 
```
if response.status_code != 404:
    parsed = response.json()
    return redirect(parsed['target_url'])
```

[note `permanent` is `False` by default](https://docs.djangoproject.com/en/2.2/topics/http/shortcuts/#django.shortcuts.redirect)

agiamas: this would try to parse a response in 500 and other error response status codes...

Thought that this redirect() is our own method, but since I see it's standard library, I dont think its gonna change without us noticing..

core/middleware.py
agiamas: done

core/middleware.py
agiamas: done

core/middleware.py
agiamas: done and commented on slack

PR #271, opened 2019-08-13
Put capital invest option in triage behind feature flag

jessgilbert:

 - [x] Changelog entry added.
 - [x] (if adding env vars) Added any new environment variable to vault.


contact/tests/test_forms.py
```diff
@@ -308,3 +308,11 @@ def test_selling_online_overseas_business_invalid_form():
...
     assert form.is_valid() is False
     assert form.errors == {'company_number': ['This field is required.']}
+
+
+@pytest.mark.parametrize('value', (True, False,))
+def test_routing_forms_capital_invest_feature_flag(value, feature_flags):
+    feature_flags['CAPITAL_INVEST_CONTACT_IN_TRIAGE_ON'] = value
+    choices = forms.InternationalRoutingForm().fields['choice'].choices
+
+    assert any(value == constants.CAPITAL_INVEST for value, label in choices) is value
```
richtier: could be simplified to 
```py
assert constants.CAPITAL_INVEST in [value for value, _ in choices]
````

File: contact/forms.py [hunk 2]
```diff
@@ -50,6 +50,16 @@ def __init__(self, *args, **kwargs):
...
 
 
+class CapitalInvestContactInTriageFeatureFlagMixin:
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if not settings.FEATURE_FLAGS['CAPITAL_INVEST_CONTACT_IN_TRIAGE_ON']:
+            self.fields['choice'].choices = [
+                (value, label) for value, label in self.CHOICES
+                if value != constants.CAPITAL_INVEST
+            ]
```
richtier: we use this approach in three places in this file. may be worth creating a reusable mixin

PR #279, opened 2019-08-16
Xot 1007 search feedback content

jeanettemariemurphy:

 - [x] Changelog entry added.


search/forms.py
```diff
@@ -26,8 +25,7 @@ class FeedbackForm(ZendeskAPIForm):
     from_search_page = IntegerField(
         widget=HiddenInput(), required=False)
     contactable = forms.ChoiceField(
-        label=('May we contact you with'
-               ' follow-up questions about your experience?'),
+        label=('May we contact you with some brief follow-up questions on your experience?'),
```
richtier: as it's all on one line we can remove the parans :)

jeanettemariemurphy: done :)

PR #285, opened 2019-08-20
Fixed feature flag mixin and added tests

jessgilbert:

 - [x] Changelog entry added.

Both off:
image

Both on:
image

Just capital invest off:
image

Just exporting to the uk off:
image

### DIFF
--- a/contact/forms.py
+++ b/contact/forms.py
@@ -45,7 +45,7 @@ class ExportingToUKOptionFeatureFlagMixin:
         super().__init__(*args, **kwargs)
         if not settings.FEATURE_FLAGS['EXPORTING_TO_UK_ON']:
             self.fields['choice'].choices = [
-                (value, label) for value, label in self.CHOICES
+                (value, label) for value, label in self.fields['choice'].choices
                 if value != constants.EXPORTING_TO_UK
             ]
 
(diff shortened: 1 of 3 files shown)


contact/forms.py
```diff
@@ -45,7 +45,7 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         if not settings.FEATURE_FLAGS['EXPORTING_TO_UK_ON']:
             self.fields['choice'].choices = [
-                (value, label) for value, label in self.CHOICES
+                (value, label) for value, label in self.fields['choice'].choices
```
MichaelWalker: This pattern of continuously mutating a field makes me a bit nervous, because it's very side-effect-y.

It has the effect of making it very difficult to understand what values the field is actually going to have at any given time, and as a result it makes bugs (like the one you are trying to fix here) much more likely.

(ie - To understand just this field, I actually need to look through and understand 4 classes!!)

So it would be nicer if we could find another way to do the same thing.
Is there a reason to do all of this by adding a bunch of mixins? Would it be simpler to just do this in a single place in the base form?

Perhaps we should have a chat and see if we can come up with an improvement! :)

richtier: we can indeed improve :)

as this issue is already present I think this is not blocking merge (the prod deployment today relies on it) but sure a follow-up PR would be well received. Due to today's deadline, unfortunately, we're going with "done is better than perfect"

PR #287, opened 2019-08-20
Tidied up picking choices from feature flag

jessgilbert:

 - [x] Changelog entry added.

### DIFF
--- a/contact/forms.py
+++ b/contact/forms.py
@@ -40,36 +40,6 @@ INDUSTRY_CHOICES = (
 )
 
 
-class ExportingToUKOptionFeatureFlagMixin:
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


MichaelWalker:
Looks good to me! I was quite closely involved with the writing of it though, so I'll let @richtier take a look before approving!

contact/forms.py
```diff
@@ -148,58 +118,70 @@ class ExportOpportunitiesRoutingForm(forms.Form):
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields['choice'].choices = international_choices()
+
     choice = forms.ChoiceField(
         label='',
         widget=forms.RadioSelect(),
-        choices=CHOICES,  # possibly updated by mixin
+        choices=[],
```
richtier: a comment would helpful to the idle developer  :)
```
        choices=[],    # set in  __init__

```

PR #298, opened 2019-08-23
Gtransform 338 add external link tag and directory components update

jeanettemariemurphy:

 - [x] Changelog entry added.

### DIFF
--- a/article/templates/article/article_detail_content.html
+++ b/article/templates/article/article_detail_content.html
@@ -1,5 +1,5 @@
 {% load parse_date from article_tags %}
-{% load add_export_elements_classes breadcrumbs from directory_components %}
+{% load add_export_elements_classes add_href_target breadcrumbs from directory_components %}
 
 
 <div class="container">
@@ -53,7 +53,7 @@
         {% endif %}
         <section class="grid-row">
           <div class="column-seven-ninths article-content">
-            {{ page.article_body_text|add_export_elements_classes }}
+            {{ page.article_body_text|add_export_elements_classes|add_href_target:request|safe  }}
           </div>
         </section>
       </div>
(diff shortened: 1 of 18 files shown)


article/templates/article/article_detail_content.html
```diff
@@ -1,5 +1,5 @@
 {% load parse_date from article_tags %}
-{% load add_export_elements_classes breadcrumbs from directory_components %}
+{% load add_export_elements_classes add_href_target breadcrumbs from directory_components %}
```
MichaelWalker: Is the intention of this tag to cause external links to open in a new tab?

Is there any chance to change the name to make its intention a little more obvious (I might be a bit late to suggest that - sorry)

article/templates/article/country_guide.html
```diff
@@ -1,6 +1,6 @@
 {% extends 'article/base.html' %}
 {% load static %}
-{% load breadcrumbs card image_with_caption cta_link cta_card add_export_elements_classes convert_headings_to override_elements_css_class statistics_card_grid from directory_components %}
+{% load breadcrumbs card image_with_caption cta_link cta_card add_export_elements_classes convert_headings_to override_elements_css_class statistics_card_grid add_href_target from directory_components %}
```
MichaelWalker: We don't seem to use this import.

article/templates/article/article_detail_content.html
jeanettemariemurphy: Yes, this tag was already existing, originally from directory-ui-supplier, but as far as I could see it wasn't used before I moved it to directory components, so could rename it.  I could look at this as an enhancement if ok?

MichaelWalker: yeah of course - not really anything to do with this PR, so shouldn't hold it up! Just saw it and didn't have any idea what it did from the name! 😊

PR #299, opened 2019-08-27
Export vouchers form

richtier:

 - [x] Changelog entry added.
 - [x] Change is accessible to the standards we subscribe to.
 - [x] Added any new environment variable to vault.
 - [x] Cleaned up old flags
 - [x] Add a printscreen to the PR

image

image

image


conf/urls.py
```diff
@@ -663,6 +663,20 @@
     ),
 ]
 
+export_vouchers_urls = [
+    url(
+        r'^export-vouchers/$',
+        skip_ga360(contact.views.ExportVoucherFormView.as_view()),
```
MichaelWalker: How come we don't want analytics for these pages?

contact/templates/contact/export-voucher-form.html
```diff
@@ -0,0 +1,45 @@
...
+            <a href="/">great.gov.uk</a>
+        {% endbreadcrumbs %}
+    </div>
+
+    <div class="container">
+        <h1 class="heading-large">[Title]</h1>
+        <p>[Body]</p>
+        <form method="post" class="width-half">
+            {% csrf_token %}
+            {% render_form form %}
```
MichaelWalker: Can this just be `{{ form }}` now?

contact/templates/contact/export-voucher-form.html
```diff
@@ -0,0 +1,45 @@
...
+    <script src="{% static 'js/company-lookup.js' %}"></script>
+    <script type="text/javascript">
+        var companyName = document.getElementById('id_company_name');
+        var companiesHouseInput =  document.getElementById('id_companies_house_number');
+        function createCompaniesHouseLookup() {
+          $(document.body).on(
+            "click.SelectiveLookupCloseAll",
+            GOVUK.components.SelectiveLookup.closeAll
+          );
+          return new GOVUK.components.CompaniesHouseNameLookup(
```
MichaelWalker: It doesn't look like we ever use the result of this return.

contact/templates/contact/export-voucher-form.html
```diff
@@ -0,0 +1,45 @@
...
+        </form>
+    </div>
+{% endblock %}
+
+{% block body_js %}
+    <script src="{% static 'js/company-lookup.js' %}"></script>
+    <script type="text/javascript">
+        var companyName = document.getElementById('id_company_name');
+        var companiesHouseInput =  document.getElementById('id_companies_house_number');
+        function createCompaniesHouseLookup() {
```
MichaelWalker: We seem to only ever call this function directly after defining it.
Could we just do all of this without a function?

contact/templates/contact/export-voucher-form.html
```diff
@@ -0,0 +1,45 @@
...
+          $(document.body).on(
+            "click.SelectiveLookupCloseAll",
+            GOVUK.components.SelectiveLookup.closeAll
```
MichaelWalker: What are these things? We seem to do very similar things elsewhere, so it seems like the right thing to do, but I don't really know what it is! :)

conf/urls.py
richtier: are there sane defaults we can use? The export voucher people have not specified what to use

contact/templates/contact/export-voucher-form.html
richtier: once we update `directory-components` here it can be :)

ah I may as well do it in this PR :+1:

contact/templates/contact/export-voucher-form.html
richtier: copy and pasted. will remove sure

contact/templates/contact/export-voucher-form.html
richtier: :+1:

contact/templates/contact/export-voucher-form.html
richtier: if the dropdown is open and the user clicks elsewhere on the screen this causes the dropdown to close

conf/urls.py
MichaelWalker: not really. In other places we've been taking the approach of 'making up something sensible', and going with that.

If we need to change it later for some reason then we can do, but better to get the data than to miss it!

richtier: cool cool will do that

MichaelWalker: as for the definitions of things

`business_unit` - is the area of the business probably `domestic` in this case.
`site_section` - is the area of the website - perhaps `export` (match other similar pages if they exist)
`site_subsection` - is the section within the `site_section` - perhaps something like `voucher`

contact/templates/contact/export-voucher-form.html
MichaelWalker: 👍

PR #304, opened 2019-08-30
Xot 920 create css template

jeanettemariemurphy:

 - [x] Changelog entry added.

screencapture-exred-trade-great-8007-advice-case-study-template-2019-08-30-15_13_38


article/templates/article/article_detail_content.html
```diff
@@ -45,19 +45,26 @@ <h2 class="heading-small">Tags</h2>
 
       </div>
     </header>
-
     <div class="grid-row">
       <div class="column-three-quarters">
         {% if page.article_image %}
           <img src="{{ page.article_image.url }}" alt="" class="width-full">
         {% endif %}
+        {% if page.article_video %}
```
eadpearce: here if the page has a video we want to show it instead of the image so it's always one or the other but not both with video taking priority

article/templates/article/article_related_content_footer.html
```diff
@@ -0,0 +1,27 @@
+<section class="padding-vertical-60 background-stone-30" id="article_related_content_footer">
+    <div class="container">
+        <div class="grid-row">
+            <div class="column-three-quarters">
```
eadpearce: may worth making a nine twelfths column class instead of having to nest 7/9 inside 3/4 here

article/templates/article/article_related_content_list.html
```diff
@@ -1,15 +1,29 @@
 
+<div class="column-quarter related-content-list">
+{% if page.cta_title or page.cta_link_label and page.cta_link %}
+<div class="panel-vertical-narrow content-list margin-bottom-90">
```
eadpearce: can you indent this just one level so it's clear it's a child of the above div

PR #311, opened 2019-09-02
Fix missing answer to business type

web-bert:
Fix missing answer to the business type question on the summary page. If other was selected also fix bug that did not show this or the text entered.
Ensure all the questions are in the same order and with the same text.


web-bert:
@agiamas changelog updated

PR #323, opened 2019-09-09
Gtransform 238 non descriptive page titles

jeanettemariemurphy:

 - [x] Changelog entry added.

### DIFF
--- a/contact/templates/contact/routing/step-great-account.html
+++ b/contact/templates/contact/routing/step-great-account.html
@@ -1,6 +1,7 @@
 {% extends 'contact/wizard-domestic.html' %}
 
 {% load static from static %}
+{% block head_title %}great.gov.uk account - great.gov.uk{% endblock %}
 
 {% block step_title_container %}
     <div id="form-step-body-text">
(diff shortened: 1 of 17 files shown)


contact/templates/contact/routing/step-great-account.html
```diff
@@ -1,6 +1,7 @@
 {% extends 'contact/wizard-domestic.html' %}
 
 {% load static from static %}
+{% block head_title %}great.gov.uk account - great.gov.uk{% endblock %}
```
eadpearce: I feel like we could probably remove the first 'great.gov.uk' here?

contact/templates/contact/routing/step-great-services.html
```diff
@@ -2,6 +2,8 @@
 
 {% load static from static %}
 
+{% block head_title %}great.gov.uk account and services support - great.gov.uk{% endblock %}
```
eadpearce: same here

contact/templates/contact/routing/step-great-account.html
jeanettemariemurphy: yes, agree, it was just that it is in the page title and so following the same pattern as was agreed.  I will remove but if we do automate we will get this issue again probably

PR #324, opened 2019-09-09
Gtransform 238 non descriptive page titles

jeanettemariemurphy:

 - [x] Changelog entry added.

### DIFF
--- a/contact/templates/contact/routing/step-location.html
+++ b/contact/templates/contact/routing/step-location.html
@@ -2,7 +2,7 @@
 
 {% load static from static %}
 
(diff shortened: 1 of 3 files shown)


contact/templates/contact/routing/step-location.html
```diff
@@ -2,7 +2,7 @@
 
 {% load static from static %}
 
-{% block head_title %}Contact the Department for International Trade (DIT) for questions about trade and investment - great.gov.uk{% endblock %}
+{% block head_title %}Contact us - great.gov.uk{% endblock %}
```
MichaelWalker: haha nice change! much simpler! 😂

jeanettemariemurphy: 😂

PR #325, opened 2019-09-10
Gtransform 213 unlabelled form elements

jeanettemariemurphy:

 - [x] Changelog entry added.

### DIFF
--- a/finance/forms.py
+++ b/finance/forms.py
@@ -90,9 +90,9 @@ class HelpForm(forms.Form):
     comment = forms.CharField(
         label='Tell us about your export experience, including any challenges you are facing.',
         help_text=(
-            'We’re particularly interested in the markets you '
-            'have exported to and whether you have already '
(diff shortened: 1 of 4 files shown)


finance/forms.py
```diff
@@ -88,13 +88,13 @@ class HelpForm(forms.Form):
...
-        label='',
-        help_text='Tell us about your export experience, including any '
-                  'challenges you are facing. We’re particularly '
-                  'interested in the markets you have exported to '
-                  'and whether you have already spoken to your '
-                  'bank or a broker.',
-        widget=Textarea,
+        label='Tell us about your export experience, including any challenges you are facing.',
+        help_text=(
+            'We’re particularly interested in the markets you '
```
richtier: funky apostrophe - more consistent would be to use `’`

jeanettemariemurphy: changed

PR #332, opened 2019-09-17
Fix anon users unable to view feedback page

richtier:
 - [x] Changelog entry added.


core/mixins.py
```diff
@@ -120,25 +120,26 @@ def get_form(self, step=None, *args, **kwargs):
...
     def guess_given_name(self):
-        if self.company_profile and self.company_profile['postal_full_name']:
-            name = self.company_profile['postal_full_name']
-            return name.split(' ')[0]
+        if self.request.user.is_authenticated:
+            if self.request.user.first_name:
+                return self.request.user.first_name
+            elif self.request.user.company and self.request.user.company['postal_full_name']:
+                name = self.request.user.company['postal_full_name']
+                return name.split(' ')[0]
```
MichaelWalker: Just checking...
Does the postal full name not start with the title? 
ie, mine wouldn't be `Mr Michael Walker` right?

richtier: you're right - hence the function name is "guess" :)

the user will have an opportunity to change the value before they submit. this is used to prepopulate

PR #341, opened 2019-09-19
Add user full name from new registration journey to SOO contact form

WillTaylor22:
 - [x] Changelog entry added.

### DIFF
--- a/sso/models.py
+++ b/sso/models.py
@@ -12,10 +12,8 @@ class SSOUser(directory_sso_api_client.models.SSOUser):
         return helpers.company_profile_retrieve(self.session_id)
 
     @cached_property
-    def full_name(self):
-        full_name = None
-        if self.get_full_name() and self.get_full_name() != '':
-            full_name = self.get_full_name()
-        elif self.company and 'postal_full_name' in self.company:
+    def get_full_name(self):
+        full_name = super().get_full_name()
+        if not full_name and self.company and 'postal_full_name' in self.company:
             full_name = self.company['postal_full_name']
         return full_name
--- a/sso/models.py
+++ b/sso/models.py
@@ -13,6 +13,6 @@ class SSOUser(directory_sso_api_client.models.SSOUser):
 
     def get_full_name(self):
         full_name = super().get_full_name()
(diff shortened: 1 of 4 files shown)


sso/models.py
```diff
@@ -10,3 +10,12 @@ class SSOUser(directory_sso_api_client.models.SSOUser):
     @cached_property
     def company(self):
         return helpers.company_profile_retrieve(self.session_id)
+
+    @cached_property
+    def full_name(self):
```
richtier: let's override the existing `get_full_name` method as it's a more "djangoistic" solution
```
def get_full_name():
    full_name = super().get_full_name()
    if not full_name and self.company:
        full_name = self.company['postal_full_name']
    return full_name
```
the user is not guaranteed to have a company

WillTaylor22: Good point about it being djangoistic - implemented!

sso/models.py
```diff
@@ -10,3 +10,10 @@ class SSOUser(directory_sso_api_client.models.SSOUser):
     @cached_property
     def company(self):
         return helpers.company_profile_retrieve(self.session_id)
+
+    @cached_property
```
richtier: no need to use cached_property here - cached property is for when the method is performing an expensive operation such as doing an API call and we're happy to cache the result.

goold old simple function, please :)

sso/models.py
```diff
@@ -10,3 +10,10 @@ class SSOUser(directory_sso_api_client.models.SSOUser):
...
+    def get_full_name(self):
+        full_name = super().get_full_name()
+        if not full_name and self.company and 'postal_full_name' in self.company:
```
richtier: if `company` is present then `postal_full_name` will be in it. `if not full_name and self.company` is enough :)

WillTaylor22: This doesn't seem very defensive?

sso/models.py
WillTaylor22: Done

sso/models.py
WillTaylor22: Anyway, removed...

PR #344, opened 2019-09-20
Added redirect from international contact triage to new great-int url

jessgilbert:

 - [x] Changelog entry added.
 - [x] (if adding env vars) Added any new environment variable to vault.


conf/env/dev
```diff
@@ -58,3 +58,19 @@ LANGUAGE_COOKIE_DOMAIN=.trade.great
...
+DIRECTORY_FORMS_API_SENDER_ID=debug
+EU_EXIT_ZENDESK_SUBDOMAIN=debug
+CONTACT_EVENTS_AGENT_EMAIL_ADDRESS=debug
+CONTACT_DSO_AGENT_EMAIL_ADDRESS=debug
+CONTACT_DIT_AGENT_EMAIL_ADDRESS=debug
+CONTACT_INTERNATIONAL_AGENT_EMAIL_ADDRESS=debug
+CONTACT_ENQUIRIES_AGENT_EMAIL_ADDRESS=debug
+CONTACT_DEFRA_AGENT_EMAIL_ADDRESS=debug
+CONTACT_BEIS_AGENT_EMAIL_ADDRESS=debug
+ACTIVITY_STREAM_API_SECRET_KEY=debug
```
MichaelWalker: a couple of these should be in the secrets one rather than the general one (anything with API_KEY or SECRET etc in it)

conf/env/dev
```diff
@@ -58,3 +55,16 @@ LANGUAGE_COOKIE_DOMAIN=.trade.great
...
+CONTACT_EVENTS_AGENT_EMAIL_ADDRESS=debug
+CONTACT_DSO_AGENT_EMAIL_ADDRESS=debug
+CONTACT_DIT_AGENT_EMAIL_ADDRESS=debug
+CONTACT_INTERNATIONAL_AGENT_EMAIL_ADDRESS=debug
+CONTACT_ENQUIRIES_AGENT_EMAIL_ADDRESS=debug
+CONTACT_DEFRA_AGENT_EMAIL_ADDRESS=debug
+CONTACT_BEIS_AGENT_EMAIL_ADDRESS=debug
+ACTIVITY_STREAM_API_ACCESS_KEY=debug
+ACTIVITY_STREAM_API_URL=debug
+ACTIVITY_STREAM_API_IP_WHITELIST=debug
```
richtier: these are all in the secrets template. run `make secrets` to generate the secrets file.

the reason the email values are not hard-coded in `dev` is that would prevent devs from testing emails locally - they would have to edit `dev` and run the risk of committing personal details to github

conf/env/secrets-template
```diff
@@ -42,3 +42,15 @@ ACTIVITY_STREAM_API_SECRET_KEY=debug#DO NOT ADD SECRETS TO THIS FILE
...
 #DO NOT ADD SECRETS TO THIS FILE
 CMS_SIGNATURE_SECRET=debug#DO NOT ADD SECRETS TO THIS FILE
+#DO NOT ADD SECRETS TO THIS FILE
+ACTIVITY_STREAM_API_SECRET_KEY=debug#DO NOT ADD SECRETS TO THIS FILE
+#DO NOT ADD SECRETS TO THIS FILE
+DIRECTORY_FORMS_API_API_KEY=debug#DO NOT ADD SECRETS TO THIS FILE
+#DO NOT ADD SECRETS TO THIS FILE
+COMPANIES_HOUSE_API_KEY=debug#DO NOT ADD SECRETS TO THIS FILE
+#DO NOT ADD SECRETS TO THIS FILE
+SECRET_KEY=debug#DO NOT ADD SECRETS TO THIS FILE
```
richtier: many of these values are already in this file

conf/env/test
```diff
@@ -36,3 +36,6 @@ ACTIVITY_STREAM_API_URL=http://test.com
 ACTIVITY_STREAM_API_IP_WHITELIST=debug
 INTERNAL_CH_BASE_URL=http://search.com
 INTERNAL_CH_API_KEY=debug
+SECRET_KEY=debug
+API_SIGNATURE_SECRET=debug
+SSO_SIGNATURE_SECRET=debug
```
richtier: all these are already set in `dev`.

PR #367, opened 2019-09-25
Xot 1042 market landing page

jeanettemariemurphy:

 - [x] Changelog entry added.

screencapture-exred-trade-great-8007-markets-2019-09-25-17_41_59


content/templates/content/markets_landing_page.html
```diff
@@ -18,3 +18,24 @@
...
+    {% pagination pagination_page=pagination_page %}
+  </div>
+</section>
+{% endblock %}
+
+{% block footer_section %}
+<section id="markets-landing-footer-section" class="topic-list-section padding-top-60 padding-bottom-60">
+  <div class="container font-medium wrap-tight">
+    <div class="informative-banner">
+   <h3>Looking for specific market tarrifs and regulations?</h3> <p><a href="#" class="link">Go to market access database</a></p>
```
eadpearce: i think this should be an `h2` but i may be wrong. can you check?

PR #372, opened 2019-09-30
XOT-1042 remove duplicated contianer div

jeanettemariemurphy:

 - [x] Changelog entry added.


content/templates/content/markets_landing_page.html
```diff
@@ -4,7 +4,7 @@
 
 {% block intro_section %}
 <section id="markets-banner-section" class="topic-list-section background-stone-30 padding-top-60 padding-bottom-0">
-  <div class="container">
```
MichaelWalker: Does this still look right if the teaser is present?
I'd expect that to want to be in a container.

jeanettemariemurphy: thanks good spot, sorted now

PR #377, opened 2019-10-02
add profile phone number to contact form

WillTaylor22:
 - [x] Changelog entry added.

### DIFF
--- a/sso/models.py
+++ b/sso/models.py
@@ -13,12 +13,12 @@ class SSOUser(directory_sso_api_client.models.SSOUser):
 
     def get_full_name(self):
         full_name = super().get_full_name()
-        if full_name in ['None None', '', None] and self.company:
+        if full_name in ['', None] and self.company:
             full_name = self.company['postal_full_name']
         return full_name
 
     def get_mobile_number(self):
         mobile_number = self.mobile_phone_number
-        if mobile_number in ['None None', '', None] and self.company:
+        if mobile_number in ['', None] and self.company:
             mobile_number = self.company['mobile_number']
         return mobile_number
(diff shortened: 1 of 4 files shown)


sso/models.py
```diff
@@ -16,3 +16,9 @@ def get_full_name(self):
         if full_name in ['None None', '', None] and self.company:
             full_name = self.company['postal_full_name']
         return full_name
+
+    def get_mobile_number(self):
+        mobile_number = super().mobile_phone_number
```
richtier: no need to call super here :) 

call super if you've overridden the method you're trying to call.

this can be `self.mobile_phone_number`

WillTaylor22: Thx!

sso/models.py
```diff
@@ -16,3 +16,9 @@ def get_full_name(self):
...
+        if mobile_number in ['None None', '', None] and self.company:
```
richtier: how would the number be `'None None'`? I suspect this is copy and paste artefact we can remove :)

MichaelWalker: is `'None None'` correct? Looks a bit surprising.

WillTaylor22: I had thought someone else had added this! Removed...

WillTaylor22: I had thought someone else had added this! Removed...

sso/models.py
```diff
@@ -13,6 +13,12 @@ def company(self):
 
     def get_full_name(self):
         full_name = super().get_full_name()
-        if full_name in ['None None', '', None] and self.company:
+        if full_name in ['', None] and self.company:
             full_name = self.company['postal_full_name']
         return full_name
+
+    def get_mobile_number(self):
```
MichaelWalker: can mobile_number just be a property with a getter?

WillTaylor22: :)

PR #378, opened 2019-10-11
XOT-1137 homepage redesign update

jeanettemariemurphy:

 - [x] Changelog entry added.

screencapture-exred-trade-great-8007-2019-10-11-16_34_03

screencapture-exred-trade-great-8007-2019-10-11-16_36_17


core/forms.py
```diff
@@ -9,10 +9,10 @@ class CompaniesHouseSearchForm(django_forms.Form):
...
         label='Sector',
-        choices=[],
+        choices=[('', 'To see possible markets - select your sector')],
     )
 
     def __init__(self, sector_list, *args, **kwargs):
         super().__init__(*args, **kwargs)
         sorted_sectors = sorted(sector_list, key=lambda x: x['name'])
-        self.fields['sector'].choices = [(tag['id'], tag['name']) for tag in sorted_sectors]
+        self.fields['sector'].choices += [(tag['id'], tag['name']) for tag in sorted_sectors]
```
richtier: I'm concerned that this will update choices multiple times:

user 1's request: adds `sorted_sectors` sectors to `choices`
user 2's request: also adds `sorted_sectors` sectors to `choices`
user 3's request: also adds `sorted_sectors` sectors to `choices`

:)

please avoid this possibility by doing something like:
```
SECTOR_CHOICES_BASE = [('', 'To see possible markets - select your sector')]
...
sector = forms.ChoiceField(
    ...
    choices=SECTOR_CHOICES_BASE,
...
def __init__(self, sector_list, *args, **kwargs):
    self.fields['sector'].choices = SECTOR_CHOICES_BASE +  [(tag['id'], tag['name']) for tag in sorted_sectors]

```

for context the reason I'm concerned is I caused this very problem before :laughing:

jeanettemariemurphy: thanks @richtier - is this ok now?

PR #404, opened 2019-10-29
Fix conf/urls

WillTaylor22:
Not sure this is worthy of a changelog entry.

This is re-adding some code that seems to have been accidentally deleting in another dev's commit, breaking the app.

### DIFF
--- a/conf/urls.py
+++ b/conf/urls.py
@@ -11,6 +11,7 @@ from django.contrib.auth.decorators import login_required
 from django.contrib.sitemaps.views import sitemap
 from django.urls import reverse_lazy
 from django.views.generic import TemplateView
+from django.views.generic.base import RedirectView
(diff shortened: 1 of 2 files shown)


conf/urls.py
```diff
@@ -11,6 +11,7 @@
 from django.contrib.sitemaps.views import sitemap
 from django.urls import reverse_lazy
 from django.views.generic import TemplateView
+from django.views.generic.base import RedirectView
```
richtier: hmm how did the tests not fail before this fix was in place?

https://github.com/uktrade/great-domestic-ui/blob/81ad48cef8ba4e4c9c1054785647a48cdf206191/conf/urls.py#L365

sdonk:
@WillTaylor22  Is this PR still relevant? It was opened 22 days ago

PR #420, opened 2019-11-15
XOT-1171 Add redirects

sdonk:

 - [x] Changelog entry added.


sdonk:
No tests because they don't add any value

PR #430, opened 2019-11-25
XOT-1135 - reset the SOO contact form when restarting it

WillTaylor22:
 - [x] Changelog entry added.


contact/views.py
```diff
@@ -489,6 +490,14 @@ def get_template_names(self):
     def get_form_kwargs(self, step):
         # skipping `PrepopulateFormMixin.get_form_kwargs`
         form_kwargs = super(mixins.PrepopulateFormMixin, self).get_form_kwargs(step)
+
+        # Reset form if at start
+        if step == self.CONTACT_DETAILS and \
+           self.request.path != reverse(
+               'contact-us-soo', kwargs={'step': 'finished'}
```
richtier: bit odd doing this in `get_form_kwargs`. let's chat in slack :)

PR #440, opened 2020-01-09
Xot 1224 market landing

jeanettemariemurphy:

 - [x] Changelog entry added.

screencapture-exred-trade-great-8007-markets-2020-01-09-12_19_51


conftest.py
```diff
@@ -8,12 +8,19 @@
...
+from directory_components.context_processors import urls_processor
+
 
 @pytest.fixture(autouse=True)
 def clear_django_cache():
     cache.clear()
 
 
+@pytest.fixture(autouse=True)
+def context():
```
richtier: as far as I can tell this fixture is used in only a few tests. I would instead move this to the tests:

```
context = urls_processor(None)
```

simpler :)

content/templates/content/markets_landing_page.html
```diff
@@ -2,57 +2,165 @@
...
+                      {% endfor %}
+                    </ul>
+                  </div>
+              </li>
 
-    {% if number_of_results > 0 and tag_name %}
-    <h2 class="font-medium margin-bottom-45">Best markets for <span class="bold">{{ tag_name }}</span> <span class="number-display highlight-circle background-active-blue">{{ number_of_results }}</span></h2>
+            </ul>
+            <div class="filters-sticky-footer" id="filters-sticky-footer">
+              <a href="#" class="cancel link">Cancel</a>
```
richtier: do these buttons with `href="#"` do anything?

jeanettemariemurphy: Yes, they are for the mobile view only though, they are bound to the event handlers clearMobileFilters and submitForm in the javascript.  The mobile view has the filters as a fullpage overlay, so cancel button closes that and the other submits the form.  The mobile form has a different behaviour to the desktop form in this way.

conftest.py
jeanettemariemurphy: This was added by @eadpearce originally as a fix to a components change - I merged this into my branch - would you be happy for this change be done retrospectively if it is needed?

jeanettemariemurphy: I have update this now - please can you recheck? thanks

PR #447, opened 2020-01-14
XOT-1224 remove regions sorting, add text format filter to regions fi…

jeanettemariemurphy:
…lters and mobile css tweaks

 - [x] Changelog entry added.


content/templates/content/markets_landing_page.html
```diff
@@ -64,7 +64,7 @@ <h2 class="heading-small margin-top-0 margin-bottom-30 margin-bottom-15-m" id="f
                       <li class="multiple-choice margin-bottom-0">
                         <input type="checkbox" value="{{ region.name }}" id="region_{{ region.id }}" name="region" {% if region.name in selected_regions %} checked {% endif %} />
                         <label for="region_{{ region.id }}">
-                          {{ region.name }}
+                          {{ region.name|title }}
```
eadpearce: what does this filter do?

jeanettemariemurphy: capitalises the word island in the regions - so we don't have to do on the back end.

richtier: neat :+1:

PR #462, opened 2020-01-30
Xot 1289 cookie preference settings page

jeanettemariemurphy:

 - [x] Changelog entry added.

screencapture-exred-trade-great-8007-cookies-2020-01-30-12_33_24


conf/urls.py
```diff
@@ -132,6 +132,12 @@
         core.views.AccessibilityStatementDomesticCMS.as_view(),
         name='accessibility-statement'
     ),
+    url(
+        r"^cookies/$",
+        skip_ga360(
```
jeanettemariemurphy: I am not sure if this page should be skip ga360 - i am assuming so as cookies are likely not set at this point, although they could be a returning user.

PR #464, opened 2020-01-30
No ticket change market filters onload event

jeanettemariemurphy:

 - [x] Changelog entry added.


jeanettemariemurphy:
Need to change the onload event window.onload will only fire once per page, so this was breaking the cookie banner on markets.

PR #508, opened 2020-03-31
Update sector choices in contact form

richtier:

 - [x] Changelog entry added.
 - [x] Ran the `make manage download_geolocation_data` command


contact/forms.py
```diff
@@ -52,7 +53,7 @@ def _wrapped(self):
     ('FOREIGN', 'UK branch of foreign company'),
     ('OTHER', 'Other'),
 )
-INDUSTRY_CHOICES = [('', 'Please select')] + choices.SECTORS + [('OTHER', 'Other')]
+INDUSTRY_CHOICES = [('', 'Please select')] + CONTACT_FORM_INDUSTRIES_CHOICES + [('OTHER', 'Other')]
```
jestanoff: If you are repeating `[('', 'Please select')]` and `[('OTHER', 'Other')]` for each set of choices, would be good if you move these into `core.constants` as well

PR #528, opened 2020-05-22
Staging release

richtier:


core/templates/core/request-export-support-form.html
```diff
@@ -24,7 +24,7 @@
 		<div class="container">
 			<div class="column-full">
 				<div class="exred-triage-form">
-					<form action="{% url 'marketing-join-form' %}" method="post" class="{% block form_class %}lg-6{% endblock %}">
+					<form method="post" class="{% block form_class %}lg-6{% endblock %}">
```
kowalcj0, line 27: Where is the action for this form defined then?

richtier: the default behavior is to submit the form the to current page - which is what we want :)

PR #537, opened 2020-06-17
Add consent request for feedback service

WillTaylor22:
- [X] Changelog entry added.
- [X] (if changing the UI) Add a printscreen to the PR
 
image

### DIFF
--- a/contact/forms.py
+++ b/contact/forms.py
@@ -34,7 +34,7 @@ class LazyEUExitLabel(LazyObject):
         return self.PRE_BREXIT
 
 
-MARKETING_CONSENT_LABEL = mark_safe(
+MARKETING_CONSENT_LABEL = (
     'Tick this box if you are happy to receive future communications from'
     ' the great.gov.uk service.'
 )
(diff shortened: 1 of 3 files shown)


contact/forms.py
```diff
@@ -34,6 +34,11 @@ def _wrapped(self):
         return self.PRE_BREXIT
 
 
+MARKETING_CONSENT_LABEL = mark_safe(
```
eadpearce, line 37: no need to use mark_safe here as there's no html in the label

contact/forms.py
```diff
@@ -266,6 +271,7 @@ class FeedbackForm(SerializeDataMixin, ZendeskActionMixin, forms.Form):
     )
     captcha = ReCaptchaField(label_suffix='')
     terms_agreed = forms.BooleanField(label=TERMS_LABEL)
+    marketing_agreed = forms.BooleanField(label=MARKETING_CONSENT_LABEL)
```
eadpearce, line 274: this field should be not required otherwise it'll need to be checked in order to submit the form

contact/tests/test_forms.py
```diff
@@ -22,6 +22,7 @@ def domestic_data(captcha_stub):
         'comment': 'Help please',
         'g-recaptcha-response': captcha_stub,
         'terms_agreed': True,
+        'marketing_agreed': True,
```
eadpearce, line 25: can you change one of these to false so we know the form still submits without checking the marketing box?

PR #547, opened 2020-08-28
Staging release

depsiatwal:


conf/url_redirects.py
```diff
@@ -354,6 +354,12 @@
             )
         ),
     ),
+    url(
+        r'^local-export-support/apply/$',
+        QuerystringRedirectView.as_view(
+            url='/contact/export-advice/business/',
+        ),
+    ),
```
webbyfox, line 362: Not sure, what happened here as this changes I've deployed yesterday so it should be in release branch already?

webbyfox: its; there in release https://github.com/uktrade/great-domestic-ui/blob/release/conf/url_redirects.py#L361

PR #550, opened 2020-09-11
GAA-16 - remove alt text from homepage images

lewis-coulson-dit:


core/templates/core/landing_page.html
```diff
@@ -26,7 +26,7 @@ <h2 class="heading-large margin-bottom-30">{{ page.how_dit_helps_title }}</h2>
     <div class="grid-row wrap-tight">
       {% for column in page.how_dit_helps_columns %}
           <div class="column-third-m column-full">
-            {% card title=column.text subheading=column.source url=column.url img_src=column.image.url img_alt=column.image_alt html_snippet=column.content no_padding_card=True transparent_card=True %}
+            {% card title=column.text subheading=column.source url=column.url img_src=column.image.url html_snippet=column.content no_padding_card=True transparent_card=True %}
```
webbyfox, line 29: worth to add alt here no? `alt=column.image.alt`

requirements.txt
```diff
@@ -15,7 +15,7 @@ directory-api-client==20.0.0  # via -r requirements.in
 directory-ch-client==2.1.0  # via -r requirements.in
 directory-client-core==6.1.0  # via directory-api-client, directory-ch-client, directory-cms-client, directory-forms-api-client, directory-sso-api-client
 directory-cms-client==11.1.0  # via -r requirements.in
-directory-components==35.15.2  # via -r requirements.in
+directory-components==35.22.0  # via -r requirements.in
```
webbyfox, line 18: latest is 35.23.0

core/templates/core/landing_page.html
lewis-coulson-dit: Nope the ticket is to remove the alt text on these images as they are just decorative

requirements.txt
lewis-coulson-dit: wow things move fast around here :)

core/templates/core/landing_page.html
webbyfox: with no role as accessibility will still fail, right?

lewis-coulson-dit: Well this was the advice given by accessibility auditors to remove the alt text completely

webbyfox: Also, card tag is coming `directory-components` so if you remove `image_alt` parameter here it will still produce alt with an empty string. https://github.com/uktrade/directory-components/blob/master/directory_components/templates/directory_components/card.html#L17

lewis-coulson-dit: Yup that's what we want - https://uktrade.atlassian.net/browse/GAA-16

webbyfox (approved):

PR #553, opened 2020-09-15
Fixed header tag

webbyfox:
 - [x] Changelog entry added.


stevejalim (approved):
I am trusting that H1 is the way to go (and not H2) - so LGTM!

PR #576, opened 2020-10-27
feat: support activitystream 1+2 simultaneously

samuelhwilliams:
We've (Data Infrastructure) deployed a new version of Activity Stream
running Elasticsearch 7, which has a slightly updated API. The 'total'
key is an object in ES7 and a plain int in ES6. We want to support both
activity streams for the transition so that we don't need to sync the
config change exactly.

 - [x] Changelog entry added.

### DIFF
--- a/search/helpers.py
+++ b/search/helpers.py
@@ -31,7 +31,16 @@ def parse_results(response, query, page):
         )
     else:
         results = serializers.parse_search_results(content)
+
+        # ActivityStream v1 / ElasticSearch v6 format - total is a raw int
         total_results = content['hits']['total']
+
+        # Check if we're using ActivityStream v2 (ElasticSearch v7) format by trying to grab the inner value
+        try:
+            total_results = total_results['value']
+        except TypeError:
+            pass
+
         total_pages = ceil(total_results/float(RESULTS_PER_PAGE))
 
     prev_pages = list(range(1, current_page))[-3:]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Pre-release
 
 ### Implemented enhancements
+- No ticket - added simultaneous support for Activity Stream 2 (Elasticsearch 7).
 
 ### Bugs fixed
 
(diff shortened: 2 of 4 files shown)


samuelhwilliams:
@depsiatwal hope you don't mind me tagging you - you were recommended as potentially a good reviewer for this change :)

search/helpers.py
```diff
@@ -31,7 +31,12 @@ def parse_results(response, query, page):
         )
     else:
         results = serializers.parse_search_results(content)
-        total_results = content['hits']['total']
+        try:
+            # ActivityStream v2 / ElasticSearch v7 format
+            total_results = content['hits']['total']['value']
+        except TypeError:
```
depsiatwal, line 37: I understand the rational bit not sure catching an exception is best flow for this use.
What if we set an environment feature flag (which is set in vault) for the switch over. 
Is there any reason why simultaneous support is required ?

depsiatwal (commented):
Is there a better way to support the two versions ?

Feature  Flagging for example 
What would happen if we got a genuine  value error now ?

search/helpers.py
samuelhwilliams: Happy to switch to a feature flag if you prefer, though it feels like it makes the deployment a little riskier because if we mess up the feature flag config somehow but move over to Activity Stream 2 then this code will fail because it won't have a fallback to Activity Stream 1 format.

samuelhwilliams:
> What would happen if we got a genuine value error now ?

Not totally sure what you mean by this - if you mean "what does great-domestic-ui do when a TypeError is raised when trying to parse results", it looks like it throws a 500. Possibly refactoring slightly to the code below would minimise the risk of catching a TypeError incorrectly ... This would keep behaviour identical in all cases other than when `content.hits.total` is an object.

```python
        total_results = content['hits']['total']

        # Check if we're using ActivityStream v2 (ElasticSearch v7) format by trying to grab the inner value
        try:
            total_results = total_results['value']
        except TypeError:
            pass
```

agiamas (approved):
lgtm

PR #585, opened 2020-12-16
GBAU-848 - Trade barrier page copy updates

lewis-coulson-dit:
Various copy changes to the trade barrier page.

 - [X] Changelog entry added.


webbyfox (approved):
LGTM

PR #590, opened 2020-12-21
Bump cryptography package

lewis-coulson-dit:


webbyfox (approved):

webbyfox:
@agiamas This should be ok to merge, right?

webbyfox (changes requested):
Ah, @lewis-coulson-dit  Re-reviewing this PR again. I think you need to make changes in requirements.in files and then run `make requirements`

PR #595, opened 2021-01-20
Removed EU transtion option from contact form

webbyfox:

 - [x] Changelog entry added.


bobmeredith (approved):
LGTM.
Nice spring-clean!

PR #598, opened 2021-02-04
feature: remove Aventri events from the search ElasticSearch query

abbas123456:
The ElasticSearch/Activity Stream feed for Aventri is currently disabled but will be re-enabled soon. However once it is re-enabled it will return all events from the Aventri API, in contrast to the original feed which did some filtering. This will cause undesired search results to appear on great.gov.uk if the ElasticSearch query remains as it is.

Therefore this PR removes Aventri events from the ElasticSearch query and a separate PR will be raised at a later stage to update the query with the relevant filters.

### DIFF
--- a/search/helpers.py
+++ b/search/helpers.py
@@ -145,12 +145,6 @@ def format_query(query, page):
                             'query': 'dit:Service',
                             'boost': 20000
                         }
-                    }},
-                    {'match': {
-                        'type': {
-                            'query': 'dit:aventri:Event',
-                            'boost': 10000
-                        }
                     }}
                 ],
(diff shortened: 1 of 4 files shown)


webbyfox:
Hi @abbas123456 can you please point me out the ticket for this PR?

abbas123456:
> Hi @abbas123456 can you please point me out the ticket for this PR?

https://uktrade.atlassian.net/browse/GBAU-921

webbyfox (approved):
Hi @abbas123456  Thanks for it. Can you please add ChangeLog entry as well.

PR #600, opened 2021-02-15
GBAU-918 Fix reCaptcha token timeout problem

bobmeredith:
On changing to reCapcha V3, there is now a token that expires 2 minutes after activation.  This has been causing some users to have failed form submissions.
This change overrides the default captcha implementation and activates the token, only on the user pressing the submit button - hence the token expiry will no longer be a problem.

 - [x] Changelog entry added.
 - [ ] (if hotfix) Has made PR into develop too

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Implemented enhancements
 
 ### Bugs fixed
-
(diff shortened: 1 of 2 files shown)


depsiatwal (approved):
lgtm

PR #618, opened 2021-05-07
noticket: Upgrade Django to 2.2.22

stevejalim:
Also pin pip-tools version so we don't get it downgraded and unable to run without being upgraded again

 - [X] Existing changelog entry is still appropriate
 - [X] (if there are vulnerable requirements) Upgraded any vulnerable dependencies.
 - [X] (if updating requirements) Requirements have been compiled.


depsiatwal (approved):
LGTM

PR #624, opened 2021-05-14
add redirect in great-domestic from /companion/

agiamas:
sibling ticket to be created in great-cms so that we dont lose it over post GP2 launch.
 - [X] Changelog entry added.


depsiatwal (approved):
LGTM

PR #641, opened 2021-07-05
GP2-2988 - Make SOO contact form name fields editable if not already set

stevejalim:
BUGFIX: Make the name fields in the SOO contact wizard to still be editable if the first and/or last name is not yet known

This is needed because Great V2 allows users to sign up without providing a first or last name, which would then cause the overall wizard to 500.

**NOTE THAT this fix is has been SEPARATELY applied to great-cms, for the contact that have been migrated from V1 into V2, but this PR is to get the fix out to current production faster**

Tests work at the Python level, but tested manually, too, of course.

### Workflow

- [X] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-2988
- [X] Jira ticket has the correct status.
- [X] [Changelog](CHANGELOG.md) entry added.

### Reviewing help

- [X] Includes screenshot(s)

**BEFORE** - see https://uktrade.atlassian.net/wiki/spaces/Great/pages/2681405487/SOO+applications+-+make+first+and+last+name+fields+editable

Screenshot 2021-07-05 at 11 14 11

**AFTER**

Screenshot 2021-07-01 at 18 04 15

(and if details are already present...)
Screenshot 2021-07-01 at 18 04 26

### Merging

- [X] reviewers can merge this

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - GP2-2784 - Bump directory components with header and footer updates
 
 ### Bugs fixed
+- GP2-2988 - Make SOO contact form name fields editable if not already populated.
 - GP2-2556 - Update verbose class implementation
 - GBAU-957 - redirect /companion/ to digital-companion.ava-digi.de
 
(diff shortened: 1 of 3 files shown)


bobmeredith (approved):
LGTM

